Issue #17, opened 2025-02-03
Return type of `get_delay_estimate()`


sbfnk:
Minor issue but this is currently a data frame but to me that seems slighty over-the-top infrastructure for something that could be a (0-indexed) vector.

kaitejohnson:
Agreed, I think in every subsequent step we just use the pmf

PR #18, opened 2025-02-03
Issue 16:  Estimate uncertainty

### LINKED ISSUE #16: Write function to estimate uncertainty
**Goal**
We would like to estimate the uncertainty in the point estimate of the nowcast using the already observed data. We will start by assuming a negative binomial observation model (which is consistent with our assumption of having count data). 

**Approach**
We will adopt the approach from the [RESPINOW codebase](https://github.com/KITmetricslab/RESPINOW-Hub/blob/39e2b17bc79492b0aee4c0b615a1c8dbf978ef53/code/baseline/functions.R#L142) to start, which iterates through each forecast
date and delay, and computes what would have been estimated as the point nowcast for each forecast date delay combination, and then estimates a dispersion parameter as a function of the delay (rather than a static one that applies across all delays). 

**Alternative approaches**
I previously also [tested out](https://github.com/kaitejohnson/scratch/blob/3ad11089c19bec70e2baba474e941430c62e47e1/baseline_EDA/baseline_EDA.Rmd#L273) a slightly simpler version of this, which estimated a single dispersion parameter by comparing the point nowcasts to the observations (but using future data).

kaitejohnson:
## Description

This PR closes #16, #17, and #19

This PR adds a function called `estimate_uncertainty()` which uses the reporting triangle being nowcasted and the delay distribution previously estimated to recompute nowcasts as of every time point. It then takes the expected nowcast and compares it to the actually observed nowcasted values, doing so for each delay d, except for delay 0 since this will never be nowcasted. A MLE estimator is used to estimate the dispersion (assuming negative binomially distributed observation error) at each delay d.  This involved:

- modifying the `apply_delay()` function, so it no longer requires there be more rows than columns in the reporting triangle (this way we can use it to produce a nowcast for a truncated triangle as of a certain time point). 
- adding an internal function, `replace_bottom_right_with_NA()`  that replaces the bottom right components of matrix, excluding the diagonal, with NAs, which is necessary to get the triangle as of the last row (time point) when truncating the observed reporting triangle. 
- adding an internal function, `.conditional_sum_cols()` which computes the sum across rows for a specific column for indices that are indicated with TRUE. This is used to incorporate only the components of a matrix that were both nowcasted and later observed in each iteration of the nowcast date
- adding an internal function `.fit_nb()` which fits a negative binomial using `stats::optimize()` to a vector of expectations and observations (adapted from the KIT code)
- adds basic tests for this, we will still need to add more as we define more narrowly what the user can and cannot pass in to all of the functions 

Scope creep:
- addresses #17 to make the return type of `get_delay_estimate()` a vector
- addresses #19 to add a bunch of tests to the validate functions 

Note, the code coverage report is not up to date. When I look in the `test_coverage` action it says the rate limit was reached... I am guessing it is only generated periodically rather than with every push?

## Question
It might make sense to have @jbracher review this implementation at some point, as I want to make sure that I understood both what was happening in the code (which I tried to maintain a semblance of) and what was written up. 

## Checklist


seabbs:
> doing so for each delay d, except for delay 0 since this will never be nowcasted

Why?

seabbs (commented):
It looks like this is a stacked PR that has gone a bit wrong (i.e because it targets main)

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+#'   triangle = triangle,
+#'   max_delay = 3,
+#'   n_history = 4
+#' )
+#' disp_params <- estimate_uncertainty(
+#'   triangle_to_nowcast = triangle,
+#'   delay_pmf = delay_pmf,
+#'   n_history_dispersion = 5
+#' )
+estimate_uncertainty <- function(triangle_to_nowcast,
```
seabbs, line 47: will you ever what to support estimating uncertainty from multiple timeseries that may not be well represented by this input approach? One alternative would be to pass in a list or long data frame here with many triangles and then provide tools to construct that input from your single triangle format

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+  # dispersion (get rid of early rows that we're not using and add NAs to
+  # bottom right of the triangle)
+  matr_observed <- triangle_to_nowcast[
+    (nrow(triangle_to_nowcast) - n_history_dispersion):
+    nrow(triangle_to_nowcast),
+  ] |>
+    .replace_lower_right_with_NA()
+
+  results <- lapply(seq_len(n_history_dispersion), function(t) {
+    # Truncate the matrix observed by ignoring rows after t, replace rows that
```
seabbs, line 80: as before anything in a loop should be its own function so we can test it

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  # Get the truncated matrix of observations you will use to estimate the
+  # dispersion (get rid of early rows that we're not using and add NAs to
+  # bottom right of the triangle)
+  matr_observed <- triangle_to_nowcast[
+    (nrow(triangle_to_nowcast) - n_history_dispersion):
+    nrow(triangle_to_nowcast),
+  ] |>
+    .replace_lower_right_with_NA()
```
seabbs, line 77: hmm not sure I get this part?

seabbs: i.e why do it here and in the function just below

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+    )
+    to_add <- sapply((seq_len(n_horizons) + 1),
+      .conditional_sum_cols,
+      matrix_bool1 = indices_nowcast,
+      matrix_bool2 = indices_observed,
+      matrix_to_sum = trunc_matr_observed
+    )
+    return(
+      list(
+        exp_to_add = exp_to_add,
```
seabbs, line 116: it seems like you might have more fun returning a data.frame

seabbs (commented):
I think the approach is in line with what @jbracher is doing.  I left some design notes that I think push the case towards some more overall design discussion.

kaitejohnson:
> It looks like this is a stacked PR that has gone a bit wrong (i.e because it targets main)

Yeah I wasn't sure how to go about this -- I had just been continuously merging in  `14-point-nowcast` to keep it up to date, since this relies on functionality there but is a separate issue...

R/estimate_uncertainty.R
kaitejohnson: Yes this is somewhat related to what we discussed yesterday (though I think what you are suggesting still uses the general approach of assuming partial reports). 

Other options are:
- allow user to pass in complete "reporting square" and estimate uncertainty from empirical distribution of delay pmfs

To allow for your suggestion, my guess is we would want to require the user to do something like `as_rt_list()` as an upstream step, where we would enforce that the triangles being passed in are return with a set formatting compatible with this function? 

I hadn't initially considered this, but I see how this could set us up for maximal flexibility, whereas requiring the input be a matrix of reporting triangle only might box us in. Still think we could probably punt it for now.

R/estimate_uncertainty.R
kaitejohnson: Will rewrite and add tests

R/estimate_uncertainty.R
kaitejohnson: So you don't have to do it in the `matr_observed`, however, if the `matr_observed` contains 0s rather than NAs for the bottom right missing data (related to discussion in #20 on whether we should require that missing values be NAs), then this will be treated as true observations and your expectations from `matr_observed_temp` will be compared to that. 

If we enforce that missing values should always be NAs, and 0s will be treated as observations,  then we can remove this. But as written, the other functions to estimate the delay and apply it assume that the bottom right triangle is missing.

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+    )
+
+    # We now have what we would have estimated for each delay,
+    # and the corresponding observations for those reference dates and delays
+    # (if they are present).
+    trunc_matr_observed <- matrix(matr_observed[1:t, ], nrow = t)
+    indices_nowcast <- matrix(is.na(matr_observed_temp), nrow = t)
+    indices_observed <- matrix(!is.na(trunc_matr_observed), nrow = t)
+
+    # First delay is 0, there will never be any nowcasts. We skip to
```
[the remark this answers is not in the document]
kaitejohnson: See my comment below. The method relies on a value at delay 0 to extrapolate to the other delays. Open to changing, though am not sure how we would exactly...

R/estimate_uncertainty.R
```diff
@@ -0,0 +1,213 @@
...
+  disp_params <- vector(length = n_horizons)
+  # remove rows with zero initial reports (Christmas etc)
+  to_keep <- abs(exp_to_add_already_observed[, 1]) >= 0.1
+  to_add_already_observed <- to_add_already_observed[to_keep, ]
+  # run through horizons
+  for (i in seq_along(1:n_horizons)) {
+    obs_temp <- to_add_already_observed[, i]
+    mu_temp <- exp_to_add_already_observed[, i]
+    mu_temp <- mu_temp + 0.1
+    disp_params[i] <- .fit_nb(x = obs_temp, mu = mu_temp)
```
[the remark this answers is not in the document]
kaitejohnson: Sorry I might have missed this. Your suggestion is to separate the estimate of the uncertainty into two functions 
1. creates the matrices of nowcast date and delay observations and expectations (`exp_to_add_already_observed`) and (`to_add_already_observed`) and 
2. Fit the delay d (which you could then do in parallel across delays)

R/estimate_uncertainty.R
kaitejohnson: Yeah this is a bit clunky, will rework to return a data.frame

kaitejohnson: Ok originally was thinking of just doing the rbind within the loop, but perplexity says thats not best practice. I had changed to have the function return a dataframe, but still do the rbind outside the loop

R/estimate_uncertainty.R
kaitejohnson: Did this, see https://github.com/epinowcast/baselinenowcast/blob/09fc7f241db416218759a23e69f167697103014c/R/estimate_uncertainty.R#L164 + tests

R/estimate_uncertainty.R
kaitejohnson: I think this is the main thing that still warrants some discussion in this PR. In the upstream user-facing functions (`estimate_delay()` and `apply_delay()` we are somewhat strictly requiring certain input data formats (raw reporting triangle, vector of pmf), rather than say creating an `estimate_delay_object` or `apply_delay_object`, which would be, I think, the analogous solution to what I proposed here with `as_rt_for_est_uncertainty`. 

For those upstream functions, I do think it might be overkill to require the user convert all inputs into that specific package object... 

But in this case, to allow for multiple or reporting triangles, I can see the argument for accepting a list or object as an input, and then providing an upstream function to format it in this way. 

Is it okay to have a mix of object passing and more "raw" inputs? Is that confusing to the user? 

@seabbs

R/estimate_uncertainty.R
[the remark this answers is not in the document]
seabbs: Im still not sure I understand why you cant have a report on day 0

R/estimate_uncertainty.R
seabbs: I dont really follow the object discussion here. My suggestion is in the common below and is to make the input a list or similar of data snapshots that look like what was available at the time vs creating these internally.

R/estimate_uncertainty.R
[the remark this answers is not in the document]
kaitejohnson: Ok so yes the first two are happening in [`.get_exp_and_obs_as_of_t`] (https://github.com/epinowcast/baselinenowcast/blob/abc0071fd881c8274c46cf685491ae3f6b5c1c75/R/estimate_uncertainty.R#L164), for a single time point (rather than creating a batch of retro, a batch of nowcasts, and mapping them back to one another). This more closely modelled the original code, but I see the value of the suggested approach. 

I see `estimate_uncertainty()` as the helper module/wrapper function for doing 1-3 of your bullets, but I see that as written it needs to change to allow for the individual components to also be run separately one by one. I can instead create the functions for tasks 1-3 and show how to put them together in the vignette, and then demonstrate the use of `estimate_uncertainty()`, which would be a wrapper function that takes in all of the above inputs, formats them properly, and returns an estimate of the observation error (hiding the inner components of the retrospective data and nowcasts from the user if they don't want/need it) ? 

The function to apply the observation error has not been written yet.

R/estimate_uncertainty.R
kaitejohnson: I can't think of another way that we would distinguish between 0s and true missingness either, which feels important.

R/estimate_uncertainty.R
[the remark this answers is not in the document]
kaitejohnson: As written, you can't produce a nowcast unless there is at least one value for a reference date. 

If there is only one value, it will always be on day 0. 

Therefore, you don't have an observation and nowcast on day 0, you just have an observation, even retrospectively, since we can't produce any nowcast without that starting point.

R/estimate_uncertainty.R
seabbs: I think we should resolve this out with the aim of merging this and open a issue linked to the design discussion for future refactoring

R/estimate_uncertainty.R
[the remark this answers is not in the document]
seabbs: Okay I think there is something I am missing here as to me this seems doable. Lets close this out and explore later

R/estimate_uncertainty.R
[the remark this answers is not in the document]
seabbs: I agree more planning makes sense. I can see wanting to make a wrapper around the middle two but I think the first one should be its own function for clarity (as it avoids making choices about " that takes in all of the above inputs" with no real code complexity cost). I do think all of these should be exported functions though.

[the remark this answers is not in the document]
seabbs: happy to close this out so we can merge this and refactor at a later date

R/estimate_uncertainty.R
seabbs: yes I agree

seabbs:
> Well perhaps this isn't general enough, but the method as written requires a value at delay 0 to extrapolate that reference dates values across the delays... Therefore if you run the algorithm here to compute the number expected to be added on each nowcast date at each delay, you will always get 0 (for the expected and the observed -- in neither case are you adding anything).

This sounds like where we need a method write up to discuss and zero handling to solve

kaitejohnson:
> > Well perhaps this isn't general enough, but the method as written requires a value at delay 0 to extrapolate that reference dates values across the delays... Therefore if you run the algorithm here to compute the number expected to be added on each nowcast date at each delay, you will always get 0 (for the expected and the observed -- in neither case are you adding anything).
> 
> This sounds like where we need a method write up to discuss and zero handling to solve

https://github.com/epinowcast/baselinenowcast/issues/35

seabbs:
> Yeah I wasn't sure how to go about this -- I had just been continuously

The traditional approach  would be to base this PR on that branch and then merge this into that branch before merging the whole thing. Generally best avoided if possible.

seabbs (approved):

PR #44, opened 2025-02-24
Issue 43: Write function to generate retrospective data from reporting triangle

### LINKED ISSUE #43: Write `generate_retrospective_data()` function
**Goal.**
As described in #38, this breaks up the task of estimating the uncertainty by incorporating a separate step to generate retrospective reporting triangles, which will then be used to estimate the uncertainty. 

**Solution**
- a function to take in a reporting triangle, the number of retrospective reporting triangles the user wants to use for estimating the uncertainty, and optionally the number of observations they want to use to estimate the delay from each retrospective reporting triangle, and returns a list of retrospective reporting triangles

**Alternatives**
We did consider having this done within the `estimate_uncertainty()` function, hidden from the user, but as discussed f2f decided against this to maximise modularity (see #40 )

kaitejohnson:
## Description

This PR closes #43.  Edited. 

Based on this review and f2f discussion, I am going to break this into two functions that are both external, and remove the checks for sufficient length from being embedded in this function (since these are only needed downstream when we try to generate retrospective nowcasts, we will use them there)

The two functions will be:
- `generate_retro_triangle(triangle, t)`
- `generate_retro_triangles(triangle, n_triangles)`
- adds tests for both new functions

Where `generate_retro_triangles()` simply calls `generate_retro_triangle()` `n_triangle` times, starting at the latest triangle and working up (so backwards in time t)

function_flow_no_borrow

## Checklist


R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
+#' Generate retrospective reporting triangles
+#'
+#' This function ingests a reporting triangle and the number of retrospective
+#'   reporting triangles we want to create, and iteratively generates the
+#'   reporting triangle that would have been available as of the maximum
+#'   reference time, and iteratively working backwards for
+#'   `n_history_uncertainty` snapshots
```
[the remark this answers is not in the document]
seabbs: i.e this function should just generate all the possible retrospective reporting triangles for the given input data

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'
+#' @param triangle_for_retro Matrix of the incomplete reporting triangle
```
seabbs, line 9: This is in a function for retrospective data and there are no other inputs is it helpful to add retro here or duplicative?

seabbs: does it need to be imcomplete for this function to work? I would assume not

seabbs: I would have hoped you could share the docs for this input from another function as it should be a standard input right?

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'   to be used to generate retrospective nowcasts, with rows representing the
+#'   time points of reference and columns representing the delays
+#' @param n_history_uncertainty Integer indicating the number of retrospective
```
seabbs, line 12: I don't really understand the uncertainty naming here? What is it referring to? the downstream use case?

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'   reporting triangles to be generated, always starting from the most
+#'   recent reference time. The default is to generate as many retrospective
+#'   reporting triangles that would have a sufficient number of rows to
+#'   estimate  a delay distribution and generate a nowcast from.
+#' @param n_history_delay Integer indicating the number of reference dates to be
```
seabbs, line 17: is this the right name for this? As this is just producing data it might make sense to just call it n_reference or similar?

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'   used in the estimate of the reporting delay, always starting from the most
+#'   recent reporting delay. The default here is to us the difference between
```
seabbs, line 19: I would have thought a more standard default would be to default to this equalling the number of columns (so this ends up as being the last D reports).

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+  .validate_triangle(triangle)
+
+  .validate_retro_inputs(
+    triangle,
+    n_history_uncertainty,
+    n_history_delay
+  )
+  # Will be able to remove this step if we require NAs in the bottom right
+  # of the triangle
+  matr_observed <- triangle |> .replace_lower_right_with_NA()
```
seabbs, line 60: this will get flagged at linting as shouldn't apparently do piping for a single function

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+
+  results <- lapply(seq_len(n_history_uncertainty),
+    .get_retro_matrix,
```
seabbs, line 63: if this is the singular of the wrapper 1. Don't we want it to be external so people can do retrospective nowcasts singularly. 2. Don't we want them to share names so they have the same input types/args and are called something similar.. like generate `generate_retro_triangle` and `generate_retro_triangles` or similar?

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+}
+
+
+#' Get retrospective matrix
+#'
+#' This function takes in a integer t and a reporting triangle and generates
+#'  the reporting triangle that would have been observed as of `t` days
+#'  earlier
+#'
+#' @param t Integer indicating the number of days prior to generate the
```
seabbs, line 78: not days it is just an index right

R/generate_retrospective_data.R
```diff
@@ -0,0 +1,99 @@
...
+#'  retrospective reporting triangle for
+#' @param matr_observed Matrix of the incomplete reporting triangle
+#'   to be used to generate retrospective nowcasts, with rows representing the
+#'   time points of reference and columns representing the delays
+#' @param n_history_delay Integer indicating the number of observations in
```
seabbs, line 83: I can't talk but comment punctuation is not consistent

R/validate.R
```diff
@@ -90,3 +90,44 @@
     )
   }
 }
+
+#' Validate inputs use to generate retrospective data
```
seabbs, line 94: line break between title and description

seabbs (approved):
I think functionality wise this is all good and I think the docs and tests cover it. I have some issues with the input specification and the naming but happy to be ignored

R/generate_retrospective_data.R
kaitejohnson: This is what filters to the number of datasets you need to estimate the uncertainty, it sounds like your preference is to pass back a less filtered version and remove/move all of these checks for having sufficient data to a downstream function

R/generate_retrospective_data.R
kaitejohnson: related to above with the filtering and checking to ensure you have enough data to do all of these things. Also trying to use the same parameter names for the same things, so in the `get_delay_estimte()`, we use the same variable to specify the number of observations to use to estimate the delay distribution (must be greater than or equal to the max delay)

R/generate_retrospective_data.R
kaitejohnson: Sure, I can change to that, or I can just only make the `generate_retro_triangle` function? And then I guess write a `generate_retro_triangles` that does nothing but generate multiples of the datasets from whats passed in without doing any checks for the user specs being possible (which would need to happen but in a separate step combining the wrapper functions)?

R/generate_retrospective_data.R
seabbs: I feel like the checks should be in the steps where those things happen?

seabbs: I think the difference in thought here is I am thinking about chaining methods whereas having this do the checks is a composition argument I think. My push back would be these tasks aren't really nested and so they don't need to be composed together in a joint function

R/generate_retrospective_data.R
seabbs: I think my understanding of the checks you are doing is limited but they sound like they are input checks for downstream functions vs in these utilities

R/generate_retrospective_data.R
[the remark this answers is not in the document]
kaitejohnson: See updated top-level comment, this function is now `generate_retro_triangles()` and it just generates `n_triangle` triangles, with the default being all possible triangles. The triangles are of varying number of rows, there is no filtering in this step to make them all the same length

R/generate_retrospective_data.R
kaitejohnson: As described above, this specification will be punted to the downstream function `generate_retro_nowcasts()`

R/generate_retrospective_data.R
kaitejohnson: This is updated and no longer contains these checks/specifications

R/generate_retrospective_data.R
kaitejohnson: fixed

R/generate_retrospective_data.R
kaitejohnson: Checks for downstream functionality will be incorporated in the PR for #47

R/validate.R
kaitejohnson: This was removed

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
+#' Generate retrospective reporting triangles
+#'
+#' This function ingests a reporting triangle and the number of retrospective
+#'   reporting triangles we want to create, and iteratively generates the
+#'   reporting triangle that would have been available as of the maximum
+#'   reference time, and iteratively working backwards for
+#'   `n_history_uncertainty` snapshots
```
seabbs, line 7: is this an out of date argument?

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#'
+#' @param triangle Matrix of the reporting triangle
+#'   to be used to generate retrospective nowcasts, with rows representing the
```
seabbs, line 10:
```suggestion
#'   to be used to generate retrospective triangles, with rows representing the
```

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#'   time points of reference and columns representing the delays
```
seabbs, line 11:
```suggestion
#'   time points of reference and columns representing the delays.
```

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+  if (n_triangles > (nrow(triangle) - ncol(triangle))) {
+    cli::cli_warn(
+      message = c(
+        "Not all of the triangles generated will contain sufficient ",
+        "observations to generate a nowcast from"
+      )
+    )
+  }
+
+  # Will be able to remove this step if we require NAs in the bottom right
```
seabbs, line 55: does this mean this function only works with partially observed data and you can't take a reported rectangle and generate reporting triangles from that?

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+
+  return(results)
+}
+
+
+#' Get a single retrospective triangle
+#'
+#' This function takes in a integer `t` and a reporting triangle and generates
+#'  the reporting triangle that would have been observed as of `t` units
+#'  earlier (starting from the bottom of the reporting triangle)
```
seabbs, line 72:
```suggestion
#'  earlier (starting from the bottom of the reporting triangle).
```

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#'
+#' @param t Integer indicating the number of days prior to generate the
+#'  retrospective reporting triangle for
```
seabbs, line 75:
```suggestion
#'  retrospective reporting triangle for.
```

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#' @param matr_observed Matrix of the incomplete reporting triangle
```
seabbs, line 76: Is it enforced to be incomplete? If so I think we might want a new issue to remove this constraint in the future as it seems like a constraint that isn't needed

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#'   to be used to generate retrospective nowcasts, with rows representing the
+#'   time points of reference and columns representing the delays
```
seabbs, line 78:
```suggestion
#'   time points of reference and columns representing the delays.
```

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#'
+#' @returns Matrix with `t` fewer rows than `matr_observed`, replicating what
+#'   would have been observed as of `t` days prior.
+#' @importFrom checkmate assert_integerish
```
seabbs, line 82: exported? 

examples?

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+  n_obs <- nrow(matr_observed)
+  if (t >= n_obs) {
+    cli_abort(
+      message = c(
+        "The as of time point is greater than or equal to the number of ",
+        "rows in the original triangle."
+      )
+    )
+  }
+  if (t < 0) {
```
seabbs, line 95: you can use a lower check in assert_integerish I think to do this more concisely.

R/get_delay_estimate.R
```diff
@@ -13,7 +13,7 @@
 #' @param max_delay Integer indicating the maximum delay to estimate, in units
 #'   of the delay. The default is to use the whole reporting triangle,
 #'   `ncol(triangle) -1`.
-#' @param n_history Integer indicating the number of reference dates to be
+#' @param n_history_delay Integer indicating the number of reference dates to be
```
seabbs, line 16: Can you remind me what the purpose of this name change is? It feels unrelated?

vignettes/baselinenowcast.Rmd
```diff
@@ -342,27 +342,74 @@ eventually reported (black line), but does a decent overall job of correcting
...
+
+To do this, we will first generate retrospective reporting triangles using the
+latest reporting triangle, and then we will generate nowcasts from the list
+of reporting triangles. To generate the nowcasts, we will have to specify the
+number of reporting triangles we want to use for the estimation of the
+uncertainty, which we will call `n_history_uncertainty`, as well as the number
+of observations we want to use to estimate the delay when we generate the
+nowcast, `n_history_delay`.
+
+Note, we will need to have more than `n_history_uncertainty + n_history_delay`
```
seabbs, line 369: This is a lot of text for a getting started IMO. Here you are just manually setting the default argument right? I would try and aim to keep this more minimal and signpost to the verbose explanation in the function docs for people who are interested

R/get_delay_estimate.R
seabbs: is it because of the argument naming choices you make higher up? Generally I think it is better to do naming etc at the lowest level as a self contained thing (i.e they shouldn't know what they are going to be used for) and then higher up rename and pass as needed for the context of that function. This avoids all you code depending on each other in complicated ways that differ from just function usage

vignettes/baselinenowcast.Rmd
```diff
@@ -342,27 +342,74 @@ eventually reported (black line), but does a decent overall job of correcting
...
+We'll start by generating the set of retrospective reporting triangles.
+Since we have 60 observations in our reporting triangle, we will generate 19
+reporting triangles, as we will need at least 41 observations in each to
+estimate a delay distribution with a maximum of 40.
+
+
+```{r generate-retrospective-triangles}
+retro_rts <- generate_retro_triangles(
+  triangle = triangle,
+  n_triangles = 19
```
seabbs, line 384: this is just the default right?

seabbs (dismissed):
I think this looks good. A few minor points.

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,108 @@
...
+#' This function ingests a reporting triangle and the number of retrospective
+#'   reporting triangles we want to create, and iteratively generates the
+#'   reporting triangle that would have been available as of the maximum
+#'   reference time, and iteratively working backwards for
+#'   `n_history_uncertainty` snapshots
+#'
+#' @param triangle Matrix of the reporting triangle
+#'   to be used to generate retrospective nowcasts, with rows representing the
+#'   time points of reference and columns representing the delays
+#' @param n_triangles Integer indicating the number of retrospective
```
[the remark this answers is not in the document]
kaitejohnson: I think we can change this to `n`, since the function name makes it clear I think.

There are really two numbers we want to keep track of and probably do want to distinguish between -- the number of observations (rows) to use to estimate the delay distribution (`n_history_delay`) and the number of reporting triangles to use to estimate the uncertainty (`n_history_uncertainty`). 

`n_triangles` is related to `n_history_uncertainty`, but in this stage of the function flow that has yet to be introduced.

R/generate_retrospective_data.R
seabbs: Yes. You are not using full stops consistently at the end of parameter outlines, returns, or description blocks. I can't see if there is a pattern so have just put them everywhere.

R/generate_retro_triangles.R
kaitejohnson: Yes let me fix, this should be `n_triangles`, or `n` as discussed above

R/generate_retro_triangles.R
kaitejohnson: This step actually isn't necessary anymore. If you did pass in a reported rectangle, it would still generate retrospective reporting triangles from that, with the first one assuming you hadn't observed the last row (most recent time point). 

I will remove.

R/generate_retro_triangles.R
kaitejohnson: I will edit the documentation here, it does not need to be incomplete-- the conversion from truncated matrix to reporting triangle happens within this function when `.replace_bottom_right_with_NA()` is called.

R/generate_retro_triangles.R
kaitejohnson: whoops will add example and export

R/get_delay_estimate.R
kaitejohnson: Yeah this was to distinguish between `n_history_delay` and `n_history_uncertainty` but per the discussion above I think these should be variables in the code and this should just be `n`?

vignettes/baselinenowcast.Rmd
kaitejohnson: Ok will edit. Yes exactly. Would you suggest not even specifying the argument, but just using the default? It seems important to explain the restrictions on how many rows/cols are needed in the triangle to do these things, but maybe it is just making it more confusing than only showing the default behavior

vignettes/baselinenowcast.Rmd
kaitejohnson: Yes, will remove to use default.

vignettes/baselinenowcast.Rmd
seabbs: If it were me I would just use the default and then in a sentence say this is assuming X thing and if you want to learn more look at ? docs

seabbs (commented):
As a final point I would be tempted to call this generate_triangles and generate_triangle as the input can be either a rectangle or a triangle and there is no other concept of generating a triangle that I am aware of (for which you would need the retro tag). 

This might help later as function names might get quite long. 

The counter argument to this is that it is very important to communicate the retrospective nature of this data and that using retro does this effectively

seabbs (commented):
As a final point I would be tempted to call this generate_triangles and generate_triangle as the input can be either a rectangle or a triangle and there is no other concept of generating a triangle that I am aware of (for which you would need the retro tag). 

This might help later as function names might get quite long. 

The counter argument to this is that it is very important to communicate the retrospective nature of this data and that using retro does this effectively

kaitejohnson:
> The counter argument to this is that it is very important to communicate the retrospective nature of this data and that using retro does this effectively

Yeah I see the argument for this and don't feel too strongly. I think the key thing is that retro somewhat communicates/reminds the user that what is being returned is computed iteratively from truncating the input starting at the bottom (from the most recent going backwards). For that reason, I am tempted to keep the `retro` even though it is verbose.

seabbs:
>  I think the key thing is that retro somewhat communicates/reminds the user that what is being returned is computed iteratively from truncating the input starting at the bottom (from the most recent going backwards). 

So to me this isn't communicated by this really as retrospective could be a triangle from any time. 

I think we can leave it for later thought but I would be tempted to default to shorter and see how that plays vs default to long just to be on the safe side

R/generate_retro_triangles.R
```diff
@@ -0,0 +1,119 @@
...
+#'   retrospective reporting triangles we want to create `n`, and iteratively
+#'   generates the reporting triangle that would have been available as of the
+#'   maximum reference time, working from bottom to top for `n` snapshots
+#'
+#' @param triangle Matrix of the reporting triangle/rectangle
+#'   to be used to generate retrospective triangles, with rows representing the
+#'   time points of reference and columns representing the delays.
+#' @param n Integer indicating the number of retrospective
+#'   reporting triangles to be generated, always starting from the most
+#'   recent reference time. Default is  only generate truncated matrices
```
seabbs, line 13:
```suggestion
#'   recent reference time. Default is  to only generate truncated matrices
```

R/validate.R
```diff
@@ -10,13 +10,13 @@
 #' @keywords internal
 .validate_triangle <- function(triangle,
                                max_delay = ncol(triangle) - 1,
-                               n_history = nrow(triangle)) {
+                               n_history_delay = nrow(triangle)) {
```
seabbs, line 13: I am confused why has this got longer?

seabbs (approved):

R/validate.R
kaitejohnson: fixed to `n`

seabbs:
Reviewing now. Can you reply with your decision either way and path forward?

kaitejohnson:
> > I think the key thing is that retro somewhat communicates/reminds the user that what is being returned is computed iteratively from truncating the input starting at the bottom (from the most recent going backwards).
> 
> So to me this isn't communicated by this really as retrospective could be a triangle from any time.
> 
> I think we can leave it for later thought but I would be tempted to default to shorter and see how that plays vs default to long just to be on the safe side

Went with `generate_triangles()` throughout

seabbs (approved):
Nice looks good. I think this puts us on a good path for the rest of the planned functionality in the getting started which all looks great.

PR #48, opened 2025-03-11
Issue 47: Generate retro nowcasts

### LINKED ISSUE #47: Write `generate_point_nowcasts()` function
**Goal**
As described in the updated function flow in #44, we want to write a separate function that takes in a list of incomplete reporting triangles and returns a list of completed reporting triangles (reporting squares). 

**Solution**
- write a function to `generate_point_nowcasts()` from a list (or a single matrix?) of reporting triangles. This should return a reporting triangle only of the number of rows of the `n_history_delay` specified. This will use the `get_delay_estimate()` and the `apply_delay()` function to generate the nowcast for each triangle

**Describe alternatives you've considered**
Alternatives considered were wrapping this into one `estimate_uncertainty()` function. We have decided instead to make this as maximally flexible as possible. 

**Additional context**
#38

kaitejohnson:
## Description

This PR closes #47. It should be reviewed after #44 is merged, as it depends on #44, or it can be reviewed as a target to `43-generate-retro-data`

- creates `generate_retro_nowcasts()` function which completes the reporting triangles in the list passed to it, using the specified `n_history_delay` as the number of observations (rows of the reporting triangle) to use to estimate the delay. 
- tests the functionality of `generate_retro_nowcasts()`
- adjusts the default behavior of `generate_retro_triangles()` to by default generate 1 less than was previously being generated, because the last one being generated did not contain sufficient number of rows to fill in the reporting triangle. 

## Checklist


seabbs:
I am not reviewing as asked but would it more helpful to target this at the correct branch until that one is merged? I can't really effectively skim this as the change log is so drowned out by past changes

kaitejohnson:
> I am not reviewing as asked but would it more helpful to target this at the correct branch until that one is merged? I can't really effectively skim this as the change log is so drowned out by past changes

Yes though I've just introduced a ton of conflicts so will break temporarily

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
+#' Generate retrospective nowcasts
+#'
+#' This function ingests a list of retrospective reporting triangles and
+#'   generates a list of retrospective reporting squares, or "complete"
+#'   reporting triangles. It uses the specified `n_history_delay`  number of
```
seabbs, line 5: out of date param name

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
...
+#'   observations to estimate the empirical delay for each retrospective
+#'   reporting triangle.
+#'
+#' @param list_of_rts List of reporting triangle matrices, in order
```
seabbs, line 9: what is the name of this output in estimate_triangles. Seems like this should be similar

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
+#' Generate retrospective nowcasts
```
seabbs, line 1: This is a point nowcast right? Do we want t osay that clearly somewhere?

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
...
+#'
+#' retro_rts <- generate_triangles(
+#'   triangle = triangle,
+#'   n = 2
+#' )
+#' retro_nowcasts <- generate_nowcasts(
+#'   list_of_rts = retro_rts,
+#'   n = 5
+#' )
+generate_nowcasts <- function(list_of_rts,
```
seabbs, line 44: point?

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
...
+        "at least the number of columns in the reporting triangle."
+      )
+    )
+  }
+
+
+  list_of_nowcasts <- list()
+
+  for (i in seq_along(list_of_rts)) {
+    # Estimate delay
```
seabbs, line 75: why not have a generate_nowcast function that this calls?

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
...
+    rep_tri <- list_of_rts[[i]]
+    delay_pmf <- get_delay_estimate(rep_tri, n = n)
+    # Apply delay
+    rt_complete <- apply_delay(rep_tri, delay_pmf)
```
seabbs, line 79: sorry this is a bit of a tangent but is this name a bit vague? This is actually making a point nowcast isn't it?

R/generate_nowcasts.R
```diff
@@ -0,0 +1,85 @@
+#' Generate retrospective nowcasts
+#'
+#' This function ingests a list of retrospective reporting triangles and
```
seabbs, line 3: I think I am a bit confused as to the precise role this plays which suggests to me it might need some more docs. 

It looks to me that it completes a triangle based on the delay in that triangle which is a special case of a nowcast? Do we therefore need to make that clear

seabbs (commented):
Just niggles from me. Mostly asking how much of a special case this is?

seabbs (commented):
Just niggles from me. Mostly asking how much of a special case this is?

R/generate_nowcasts.R
kaitejohnson: fixed

kaitejohnson:
> Just niggles from me. Mostly asking how much of a special case this is?

Good points. Per our f2f discussion, I have refactored this in #50 . Instead of the workflow being:
triangle -> generate_triangles -> generate_nowcasts()

It is now more modular:
triangle -> truncate_triangles -> generate_triangles -> generate_nowcasts

This functionally separate the two things `generate_triangles()` is doing: 
1. truncating the original matrix
2. replacing the bottom right with NAs 

And then `generate_nowcasts()` takes in the list of reporting triangles and generates a point nowcast based on each triangle in the list (treating each independently). I will modify the docs in #50 to make this more clear, but suggest reviewing as one given the refactor that does to this functionality

R/generate_nowcasts.R
kaitejohnson: Will go through and make sure the output names match the input names for the strings of functions

R/generate_nowcasts.R
kaitejohnson: Reworded in #50

R/generate_nowcasts.R
kaitejohnson: changed name in #50

R/generate_nowcasts.R
kaitejohnson: I think its better to have this be more modular so:
1. estimate delay
2. apply the delay to generate a nowcast

While in this instance, we are not going to apply a different delay, in other cases we could, and we could eventually modify this to instead pass in a delay from somewhere else. 

I think if we want to make a new name for `apply_delay()` we should open a new issue for it. 

I see the generation of a point nowcast as the combination of these two steps (delay estimation, and then applying that delay to fill in the triangle). That's why I chose to name the second step apply delay, but open to other suggestions

R/generate_nowcasts.R
kaitejohnson: See above

R/generate_nowcasts.R
seabbs: I think the nowcast is the bit where you adjust the data and so would say that `apply_delay` is the nowcast step. This wrapper is also nowcasting as it contains that step and its doing so by first estimating a delay and then nowcasting using it.

All that aside my main point here is that elsewhere you have singular and plural versions of this function why not have generate_nowcast that is called by generate_nowcasts?

R/generate_nowcasts.R
seabbs: can you explain it shortly here as its a bit hard to review otherwise?

seabbs:
This design shuts off the simple use of a different delay estimate in a nowcast doesn't it?

kaitejohnson:
> This design shuts off the simple use of a different delay estimate in a nowcast doesn't it?

I don't think so, the estimation of the uncertainty is separate from the generation of the "final" nowcast (we haven't written that step yet). 

This is imposing that the way uncertainty is estimated is by re-generating retrospective nowcasts using the reporting triangle passed into this chain of functions, as discussed with Johannes and in our f2f discussions. It does not allow the estimation of uncertainty from an entirely separate, and constant, delay estimate. This was my original workflow but we decided not to allow this functionality. 

The dispersion parameters that come out of this though, can be combined with a different separate delay in a future function to generate the probabilistic nowcast (it hasn't been written though!)

kaitejohnson:
Closing this as changes have been made in #50

PR #56, opened 2025-03-18
Issue 55: `generate_prob_nowcast_df()`

### LINKED ISSUE #55: Write `generate_probabilistic_nowcast_df` function
**Goal**
In alignment with the current function flow, we want to write a function that takes a list of `n_draws` expected observed reporting squares and converts this into a long form tidy dataframe, indexed by reference time `t` and delay `d`.

Image

**Solution**
- write a function `generate_probabilistic_nowcast_df(list_of_exp_obs_nowcasts)` which returns a dataframe containing the reference time and delay of each observation, as well as an indicator of whether that observation was generated as an expected observation or whether it was observed.

kaitejohnson:
## Description

This PR closes #55. It generates two functions:
- `convert_reporting_square_to_df()` which takes in a matrix that is assumed to have just observations, with rows indicating the reference time and columns indicating the delay. It converts it to a dataframe with this metadata as columns, alongside an optional draw column which tracks the sample of the expected observed reporting square (for looping)
- `nowcast_list_to_df()` takes in a list of expected observed reporting squares (nowcasts) and converts them to a long tidy dataframe indexed by the reference time, delay, and the draw
- adds tests for both functions

## Checklist


R/generate_prob_nowcasts_df.R
```diff
@@ -0,0 +1,103 @@
...
+#' list_of_exp_obs_nowcast <- add_obs_errors_to_nowcast(
+#'   comp_rep_square = point_nowcast,
+#'   disp = c(8, 1.4, 4),
+#'   n_draws = 10
+#' )
+#'
+#' nowcast_df <- generate_prob_nowcast_df(
+#'   list_of_nowcasts = list_of_exp_obs_nowcast
+#' )
+generate_prob_nowcast_df <- function(list_of_nowcasts) {
```
seabbs, line 35: it doesn't really generate following the naming logic you have used elsewhere? what about nowcast_list_to_df or similar?

kaitejohnson: yeah the other is more explicit (but could also maybe be named `rep_matrix_to_df()`?

kaitejohnson: Changed to `nowcast_list_to_df()`, left the other for now

R/nowcast_list_to_df.R
```diff
@@ -0,0 +1,108 @@
...
+
+  # Convert each matrix to a data frame with an index column
+  combined_df <- do.call(
+    rbind,
+    lapply(
+      seq_along(list_of_nowcasts),
+      function(i) {
+        return(convert_reporting_square_to_df(
+          list_of_nowcasts[[i]],
+          draw = i
```
seabbs, line 52: This assumes that the intake is a probabilistic sample rather than say a list by reference date or any other grouping. In a new issue it would be nice to think about generalising this as it isn't reflected in the docs and I think is overly restrictive.

R/nowcast_list_to_df.R
```diff
@@ -0,0 +1,108 @@
...
+#'   ),
+#'   nrow = 6,
+#'   byrow = TRUE
+#' )
+#'
+#' long_df <- convert_reporting_square_to_df(rep_square)
+#' print(long_df)
+convert_reporting_square_to_df <- function(matrix,
+                                           draw = NULL) {
+  # Convert to data.frame
```
seabbs, line 92: I don't think we need to be so over commenting obvious sterps like this (I realise LLMs love to do it).

tests/testthat.R
```diff
@@ -1,8 +1,9 @@
 library(testthat)
 library(baselinenowcast)
+library(tidyr)
```
seabbs, line 3: Rather than introducing this as a test dep just manually have the long format data in the test. This makes it easier to read/check it is okay and more stable as it doesn't introduce a hidden dep on tidyr

seabbs (dismissed):
I think that the function is overly specialised and hence needlessly brittle but this can be addressed in a follow up issue/PR

R/nowcast_list_to_df.R
kaitejohnson: Yep, I wrote this with dplyr and passed it to an LLM to do with base R then fixed bc it was wrong.... have removed.

tests/testthat.R
kaitejohnson: removed tidyr dependency and replaced with the result itself in the test

R/nowcast_list_to_df.R
kaitejohnson: Made a new issue for this, will flesh out further depending on response here. 

Perhaps the option `draw` argument should be just a more general way of tracking the items in the list, and then in the `nowcast_list_to_df` we convert that column to indicate that in this case it is a draw?

seabbs: >  returns a dataframe with a column for draw (so it does assume intake is list of draws)...

yes this is the problem

seabbs (approved):

seabbs (approved):

PR #63, opened 2025-03-24
Issue 62: write function to summarise by reference time

### LINKED ISSUE #62: Write function to summarise by reference time
**Goal**
Want a helper function that takes the long dataframe indexed by draws, reference times, and delays, and sums over all the delays within a draw to generate probabilistic nowcasts of the expected total observations at each reference time. 

**Solution**
- a function called `summarise_by_reference_time()` that ingests the long form dataframe and spits back a dataframe only indexed by reference time and draws

kaitejohnson:
## Description

This PR closes #62. It builds off of #56 so recommend reviewing that first. 

Changes made:
- adds a function called `summarise_by_ref_time()` which ingests a long tidy dataframe of draws of reporting squares and produces probabilistic nowcasts as sums across all delays 
- adds tests for this function
- modifies the "Getting Started" vignette to use this function and produce a figure as described below:

The vignette uses the new function and then joins the resulting dataframe of probabilistic nowcasts summarised by reference time with the observed data. Red line is the data as of the nowcast date (July 01, 2021), black line is the data as of 3 months later (October 01, 2021). Gray lines are probabilistic nowcasts, using a maximum delay of 40 and estimating the dispersion from 18 reporting triangles (the package defaults). 
image## Checklist


R/summarise_by_reference_time.R
```diff
@@ -0,0 +1,43 @@
...
+#' @examples
+#' nowcast_draws_df <- data.frame(
+#'   time = c(1, 1, 1, 1, 2, 2, 2, 2),
+#'   delay = c(1, 2, 1, 2, 1, 2, 1, 2),
+#'   draw = c(1, 1, 2, 2, 1, 1, 2, 2),
+#'   count = c(3, 6, 4, 7, 1, 2, 2, 3)
+#' )
+#' summary_df <- summarise_by_reference_time(nowcast_draws_df)
+#' print(summary_df)
+summarise_by_reference_time <- function(nowcast_draws_df) {
```
seabbs, line 24: I think this is a bit opaque in terms of what it actually does which is sum counts to what is reported/nowcast. I think this comes from sticking a bit closely to the summarise approach.

Traditionally what this function is putting out is the nowcast so that could be something to ground on. 

Just checking but the input here has already been fused so it is the observed data where it is available and the nowcast where it is not?

vignettes/baselinenowcast.Rmd
```diff
@@ -431,26 +432,72 @@ exp_obs_nowcasts <- add_obs_errors_to_nowcast(
 ```
 
 Next, we will take the list of expected observed reporting squares and
-convert this into a long tidy dataframe, indexed by reference time `t` and
-delay `d`, and plot the results
+convert this into a long tidy dataframe, indexed by reference time `t`,
+delay `d`, and draws.
```
seabbs, line 436: Bleed from the other PR.

vignettes/baselinenowcast.Rmd
```diff
@@ -431,26 +432,72 @@ exp_obs_nowcasts <- add_obs_errors_to_nowcast(
...
-  geom_line(aes(x = time, y = total_counts, group = draw),
-    alpha = 0.1, color = "gray", linewidth = 0.1
+# Join with the original data
+final_data <- data_long[location == "DE"][age_group == "00+"] |> # nolint
+  enw_filter_report_dates(latest_date = "2021-10-01") |>
+  enw_filter_reference_dates(
+    latest_date = "2021-07-01",
+    include_days = n_history_delay - 1
+  ) |>
+  group_by(reference_date) |>
```
seabbs, line 459: This info is in the preprocessed object and there is a exported function you can look at to get it.

That being said I think its fine to use dplyr here if you want to

seabbs (dismissed):
Overall looks good.

R/summarise_by_reference_time.R
kaitejohnson: Yes it is the observed and nowcated data fused together. 

I am not entirely sure we need this function, as it is basically just another implementation of `dply::summarise`?

Or is your suggestion to just call this something more like `get_nowcast_df()`...

vignettes/baselinenowcast.Rmd
kaitejohnson: I think I intentionally changed this bc we no longer plot the results of that particular step here...

vignettes/baselinenowcast.Rmd
kaitejohnson: Ah yes, I think I should do some getting started vignette cleaning up bc I am doing a bit of repetitive data wrangling... 

My intention was to not include too much post processing in `baselinenowcast`, which is why I am just showing how you do this here vs writing an exported package function to do this (which we also could)

Or by exported function you mean from `epinowcast`?

seabbs: yes from epinowcast

R/summarise_by_reference_time.R
kaitejohnson: changed name to `get_nowcast_df()`

vignettes/baselinenowcast.Rmd
kaitejohnson: Slight preference to just use `dplyr` since it is more likely users will be familiar will easily understand what is happening. Will address general redundancy in #68

seabbs (approved):
Approving but noting that there is now a naming collision between get_nowcast_df and get_prob_nowcast_df I think which do different things.

kaitejohnson:
> Approving but noting that there is now a naming collision between get_nowcast_df and get_prob_nowcast_df I think which do different things.

Is there? So the PR to generate_probabilistic_nowcast_df #55 actually creates two functions: `convert_reporting_square_to_df()` and `nowcast_list_to_df()` (which have the functionality of generating a probabilistic nowcast where a "nowcast" is the full reporting square)

I think the larger issue that I ran into when reorganising the methods is that I think we want to clearly define:
- reporting square with observed and imputed point estimates (I've been calling this a point nowcast)
- reporting square with observed and expected observed estimates (I've been calling this an expected observed nowcast)
- a vector of expected observed total counts at each reference time (this is being generated in this PR, so we are implicitly calling it a nowcast). 

If we call the summarised by reference time thing a nowcast, we should clearly distinguish something of that form to the reporting square objects.

Would you say a "nowcast" is typically defined as the sum of the elements across delays, and the matrices are described in the more verbose way as say reporting triangle, reporting square, etc.?

seabbs:
Ah it didn't make it in. 

> nowcast_list_to_df()

I think this is still quite a confusing conflict as it sounds like the same thing. 


> reporting square with observed and imputed point estimates (I've been calling this a point nowcast)
> reporting square with observed and expected observed estimates (I've been calling this an expected observed nowcast)

What is the difference between these two things? Without context they sound like the same thing?

seabbs (dismissed):

kaitejohnson:
> What is the difference between these two things? Without context they sound like the same thing?

One is summarised by reference time (so just indexed by reference time) e.g. could be a vector, the other is indexed by reference time and report time (a matrix)

Yeah I will make a meta-issue with a table to distinguish between these things. However, for this PR I think its reasonable to still change the name if its confusing (which I believe it is currently). What about `aggregate_by_reference_time()` instead of `get_nowcast_df()`?

kaitejohnson:
Made a new issue #74

seabbs (approved):

PR #69, opened 2025-03-28
Issue 46: refactor model definition

### LINKED ISSUE #46: Refactor the connections between model components in the methodology overview
So my original review point didn't make it for some reason. 

I think the core content here is good but the organisation between sections is quite unclear to me and I think will be very unclear to a reader. I also think the overview is really just an overview of section 2.3 and so it would be better to move it there (as that is also where the figure is mainly used). 

You could then have an overview that gives a sentence or two to each of the subsection headers and communicates how they link as that then gives an overview of the modelling approach. 

I think the structure of the sections might also need to change with more clarity on what the steps are. I think it could potentially be helpful to either first focus on the complete case and the comment on how the non-complete case works or make it clearer that the non-complete case is also producing a delay estimate that we can use in the same way. 

I think as this PR is it can be merged and I would be happy to address some/all of my concerns in a new issue/PR

_Originally posted by @seabbs in https://github.com/epinowcast/baselinenowcast/issues/37#issuecomment-2691146229_

kaitejohnson:
## Description

This PR closes #46.

It restructures the methods to be more consistent with the code (linking to the explicitly linking to the functions that pertain to the model components).

## Checklist


seabbs:
I had a go at shortening the overview: 

The baselinenowcast model, developed as a reference for the COVID-19 hospital admissions nowcasting challenge in Germany (2021-2022), utilises reporting triangles of preliminary case counts and their delays. It applies a multiplicative approach, using empirically observed historical delay distributions to estimate yet-to-be-observed cases.
Users can specify whether delay distributions are estimated from the current reporting triangle, fully observed past data, or data from different strata. The model produces point nowcasts by "filling in" the reporting triangle.
Probabilistic nowcasts are generated using a negative binomial model with means from the point nowcast and dispersion parameters estimated from past nowcast errors. These parameters can also be flexibly sourced from current data, historical data, or alternative settings.

kaitejohnson:
This looks good to me but the language here depends on our answers to #74 (because here you are using point nowcast to mean fill in the reporting triangle, which is also how I have used it, but in other places it means the estimate of the sum across delays for each reference time).

vignettes/model_definition.Rmd
```diff
@@ -18,22 +18,9 @@ vignette: >
...
-In the case where the temporal granularity of the reporting differs from the referencing, e.g.
-weekly releases of daily data, pre-processing will need to be done so that the units of the delays $d$ and the units of the reference time $t$ are the same (e.g. by filling in 0s for all the days of the week that no data is reported).
-This method is based on the method described by [Wolffram et al. 2023](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1011394).
+The `baselinenowcast` model, initially developed as a reference for the COVID-19 hospital admissions nowcasting challenge in Germany (2021-2022, [Wolffram et al](https://journals.plos.org/ploscompbiol/article?id=10.1371/journal.pcbi.1011394), utilises reporting triangles of preliminary case counts and their delays. It applies a multiplicative approach, using empirically observed historical delay distributions to estimate yet-to-be-observed cases.
+Users can specify whether delay distributions are estimated from the current reporting triangle, fully observed past data, or data from different strata. The model produces point nowcasts by "filling in" the reporting triangle.
```
kaitejohnson, line 22: I think we want to decide if a point nowcast = completed reporting square or if it is the sum of point estimates at each element in completed reporting square

seabbs (approved):
Aside from the nomenclature discussion this looks good.

seabbs:
sounds good

PR #76, opened 2025-04-07
Issue 75: Fix bug where last row is Inf

### LINKED ISSUE #75: Fix bug introducing last row as Inf in measles example
**Bug description**
Noticed when running example analysis on measles in https://github.com/epinowcast/baselinenowcast-paper/pull/9 that we are getting a row of Infs for the retrospective nowcasts. Troubleshoot to identify what is causing this and potential solutions.

kaitejohnson:
## Description

This PR closes #75. 

- `get_delay_estimate()` produces a warning if the first element of the delay_pmf is estimated at 0.
- `apply_delay()` function errors if the first element of delay_pmf is 0 and the triangle being passed in is a reporting triangle (meaning it has only a single value that is not NA in the last row).
- added tests to make sure this functionality works as expected

## Checklist


R/get_delay_estimate.R
```diff
@@ -94,5 +94,17 @@ get_delay_estimate <- function(triangle,
   # delay distribution
   pmf <- colSums(expectation) / sum(expectation)
 
+  if (pmf[1] == 0) {
+    cli_warn(
+      message =
+        c(
+          "Value of delay PMF at delay = 0 is 0. This delay PMF will not be ",
```
seabbs, line 101: I think here we may want to suggest some action a user should/can take? 

From a separation of concerns perspective do we need this warning here as having a zero is a totally valid delay and it only becomes a problem when it is used for nowcasting?

R/validate.R
```diff
@@ -100,5 +100,20 @@
...
     )
   }
+
+  # Make sure that nowcasts can be generated iteratively, ensuring delay_pmf[1]
+  # !=0 while the triangle is a reporting triangle (NAs all in bottom right)
+  if (is.na(triangle[nrow(triangle), 2]) && delay_pmf[1] == 0) {
+    cli_abort(
+      message = c(
+        "Value of delay PMF at delay = 0 is 0, and the latest reference time ",
+        "in the reporting matrix only contains a value at delay = 0. There ",
```
seabbs, line 110:
```suggestion
        "in the reporting matrix only contains a value at delay = 0. There is",
```

seabbs (approved):
This looks good. Due to wanting separation of concerns I think I would go with just erroring when the delay is applied vs doubling up but I don't feel strongly about this.

R/get_delay_estimate.R
kaitejohnson: Sure I guess its a bit overkill, can remove.

R/validate.R
```diff
@@ -100,5 +100,20 @@
       )
     )
   }
+
+  # Make sure that nowcasts can be generated iteratively, ensuring delay_pmf[1]
+  # !=0 while the triangle is a reporting triangle (NAs all in bottom right)
+  if (is.na(triangle[nrow(triangle), 2]) && delay_pmf[1] == 0) {
```
seabbs, line 106: do you want to slightly loosen this check to very nearly 0 or is this enough?

kaitejohnson: Hmmm, I think if it is very nearly 0, that means there was at least a single observation with delay =0 and then I don't think you'd want to consider reindexing

PR #84, opened 2025-04-09
74 standardise naming

### LINKED ISSUE #74: Decide on naming convention for reporting matrix vs vector of nowcasts, observed vs imputed values
**Goal**

Currently, throwing around the word nowcast to both mean the combination of observed and imputed reporting square matrix AND the vector of expected total reports at each reference time. 

We want to distinguish between:
- the reporting matrix (both triangle and square)
- vector of reports by reference time (Wolffram et al refer to this as a point nowcast)
And between:
- reporting triangle (with only observations and incomplete bottom right)
- reporting square/rectangle with observations +  imputed mean values for the bottom right elements (currently being referred to as a point nowcast)
- reporting square/rectangle with observations + expected observed values for the bottom right elements

**Context** 
https://github.com/epinowcast/baselinenowcast/pull/63#issuecomment-2766241708

**Proposed table of names**
UPDATED
| Data structure | Observations only | Mixed (obs + predictions) | Pure predictions only |
|----------------|-------------------|---------------------------|------------------------|
| Matrix format | reporting_matrix (complete)<br>reporting_triangle (incomplete) <br> incomplete_reporting_matrix | point_nowcast_matrix<br>nowcast_matrix (probabilistic) | point_pred_matrix<br>pred_matrix (probabilistic) |
| List format | reporting_matrix_list (complete)<br>reporting_triangle_list (incomplete)<br>incomplete_reporting_matrix_list | nowcast_matrix_list | pred_matrix_list |
| Vector format | observed_cases | point_nowcast<br>nowcast (probabilistic) | point_pred<br>pred (probabilistic) |
| DataFrame format | - | nowcast_df | pred_df |

Shorteners: 
`reporting --> rep`, `observed --> obs`, `incomplete --> inc`, `matrix --> mat`

kaitejohnson:
## Description

This PR closes #74. WIP. It does the following:
- modifies the model definition to add a table of names for various matrices, summarised vectors, and dataframes contaning either observations, a mix of observations, or pure predictions (see #74 for details on decisions)
- modifies model definition text to use the plain language version of these 
- modifies two functions to change their names to be consistent with this nomenclature
- modifies Getting Started vignette and documentation/variable names throughout to ensure everything is consistent. 

[Describe the changes that you made in this pull request.]

## Checklist


NAMESPACE
```diff
@@ -1,16 +1,16 @@
 # Generated by roxygen2: do not edit by hand
 
-export(add_obs_error_to_nowcast)
-export(add_obs_errors_to_nowcast)
+export(add_obs_err_to_pt_nowcast_mat)
```
seabbs, line 3: Some naming thoughts... (not required maybe for this PR

point_to_nowcast_matrix
point_to_nowcast_matrix_list 

?

NAMESPACE
```diff
@@ -1,16 +1,16 @@
...
+export(add_obs_errs_to_pt_nowcast_mat)
 export(aggregate_df_by_ref_time)
 export(apply_delay)
-export(convert_reporting_square_to_df)
+export(convert_nowcast_matrix_to_df)
```
seabbs, line 7: Inspired by posterior this could be nowcast_matrix_as_df and so on

NAMESPACE
```diff
@@ -1,16 +1,16 @@
...
 export(estimate_dispersion)
-export(generate_point_nowcast)
-export(generate_point_nowcasts)
+export(generate_pt_nowcast_mat)
+export(generate_pt_nowcast_mat_list)
 export(generate_triangles)
 export(get_delay_estimate)
-export(nowcast_list_to_df)
+export(nowcast_matrix_list_to_df)
```
seabbs, line 13: same here could be as (also either the first one shouldn't have convert orthis one should? I favour not having convert

NAMESPACE
```diff
@@ -1,16 +1,16 @@
...
 export(replace_lower_right_with_NA)
 export(truncate_triangle)
 export(truncate_triangles)
```
seabbs, line 16: is this becase the output is a list as is the input? Shoud it use a list prefix then? 

As above could be as_triangle_list?

NAMESPACE
```diff
@@ -1,16 +1,16 @@
...
+export(add_obs_err_to_pt_nowcast_mat)
+export(add_obs_errs_to_pt_nowcast_mat)
 export(aggregate_df_by_ref_time)
 export(apply_delay)
-export(convert_reporting_square_to_df)
+export(convert_nowcast_matrix_to_df)
 export(estimate_dispersion)
-export(generate_point_nowcast)
-export(generate_point_nowcasts)
+export(generate_pt_nowcast_mat)
```
seabbs, line 9: do you prefer having full generate vs shortening to gen and having point vs pt?

NAMESPACE
seabbs: with the theme of shortening could also making input implicit and call them 

add_uncertainty/dispersion or similar and add_<>_list ?

seabbs (dismissed):
Nice I think this looks good. What do you think? Is it clearer? some of the names are definitely long. 

I added a few suggestions to either better align with the scheme or to more closely track posterior. These are just ideas.

NAMESPACE
kaitejohnson: true can standardise.... have a slight preference for to as I am not a posterior user

NAMESPACE
kaitejohnson: I prefer to over as, it implies more action to me as to what is happening 

this takes a list and converts to df?

NAMESPACE
kaitejohnson: I have used pt throughout, think generate is fine but can also shorten

NAMESPACE
kaitejohnson: Oh are you referring to truncate triangles here 
This takes a single reporting triangle and returns a list of truncated reporting triangles, but the key thing its doing is truncating the original triangle

kaitejohnson:
I think it would be helpful to have the naming somewhat nailed down in this PR, to avoid deprecations in the paper workflow. Based on your review and my feedback generally I would prefer:
- we use `blank_to_blank` rather than `as_blank` (more verbose, but I think clearer) 
- we get rid of `convert` since this is redundant 
- where the function is not simply transforming the same info, we can be more explicit as to what its doing and less explicit on the inputs and outputs (e.g. keep `replace_lower_right_with_NA` bc it tells you what the function does, also pt-> draw is adding uncertainty so specify that in the name

This would require the following changes: 
`add_obs_err_to_pt_nowcast_mat` --> `add_uncertainty` with same inputs and outputs
`convert_nowcast_matrix_to_df` --> `nowcast_matrix_to_df` with same inputs and outputs
`truncate_triangles` --> `rep_tri_to_trunc_rep_mat_list` *

* Note I think this one is over complicated, I think its fine to just use the verb and rely on the arguments to indicate how the inputs and outputs are being transformed.

seabbs:
> where the function is not simply transforming the same info, we can be more explicit as to what its doing and less explicit

See my comment on the specific line. I don't think this is a good idea for the reasons given there. If you do feel strongly then we can go ahead with this approach. If we do that I think we should track these pattern breaking functions.

> we use blank_to_blank rather than as_blank (more verbose, but I think clearer)

Okay. This could get rather long. I think we only need to be this verbose if there are multiple similar functions with different inputs. Again if you feel strongly we can go with this. We can always add aliases down the line. In terms of to or as I don't think this point addresses that. Again I don't feel strongly excepting my general liking of tracking decisions others have made (i.e as in posterior).

> we get rid of convert since this is redundant

agree

> add_obs_err_to_pt_nowcast_mat --> add_uncertainty with same inputs and outputs

If this now the only function that doesn't flag its input format (i.e my point above) or is there a mix. As I said there I think this is fine as long as there is some rational for making these choices (i.e one clear one is there is no very similarlarly named function and we don't think there will be). 

> truncate_triangles --> rep_tri_to_trunc_rep_mat_list *

This is going in the reverse direction to `add_uncertainty` and again I think this is fine if there is logic. 

As I said I think my preference here would be as_reporting_triangle and as_reporting_triangle_list. I think I don't agree with the idea that the input is a triangle and it produces a matrix. Isnt our triangle format also a matrix but with NAs so this would produce a triangle and doesn't the input support both matrix and triangles? I think what we have currently is significantly better than the verbose suggestion as I see you have commented but as above I think this breaks the logic of the naming scheme with no clear pattern as to why

NAMESPACE
kaitejohnson: No -- it is taking a matrix and generating a list of `n` matrices, which have one row sliced off of the original matrix.

kaitejohnson:
> If this now the only function that doesn't flag its input format (i.e my point above) or is there a mix.
My suggestion was:
- if a function is literally just doing a transform, we name it as such `X_to_Y` where Y is an alternate form of storing the same information
- if a function is taking an action (e.g. truncating, adding uncertainty, generating a nowcast, we don't focus on inputs and outputs, we use the verb. The example of this would be `add_uncertainty()` or `truncate_triangles()`

I think `as_{output}` is less clear, even this is what posterior does. 

> As I said I think my preference here would be as_reporting_triangle and as_reporting_triangle_list. I think I don't agree with the idea that the input is a triangle and it produces a matrix. Isnt our triangle format also a matrix but with NAs so this would produce a triangle and doesn't the input support both matrix and triangles? I think what we have currently is significantly better than the verbose suggestion as I see you have commented but as above I think this breaks the logic of the naming scheme with no clear pattern as to why

It isn't just converting the reporting triangle (which is a matrix, but we are calling it a rep triangle to distinguish that it has the bottom right with NAs in the default case). It is truncating a row off of that each time and stringing them together in a list. My preference is to stick with `truncate_triangles()`

kaitejohnson:
> This was my understanding of what it is doing but in the general sense it could also do this to a matrix and hence thinking that that is a better naming scheme. It sounds like you have a set preference here so lets go with that
 
I guess I am not following the logic of why a function that truncates a triangle/matrix would just be called `as_triangle_list()` which is your suggestion if I am understanding correctly. That tells me pretty little about what its doing other than it will return a list, and notably, it will be a list of *not* what we have defined as a reporting triangle (a reporting triangle = special case of a matrix with all bottom right triangle containing NAs)

> I think we are slightly talking across each other as the first set of points here don't really seem to address my comments fully but that is okay and I think we should move on.

If you could comment with your proposed suggestions (other than for `truncate_triangles()` and the `as_` instead of `X_to_Y` that would maybe help me follow. 

As of right now I plan on making these changes:
`add_obs_err_to_pt_nowcast_mat`--> `add_uncertainty` with same inputs and outputs
`convert_nowcast_matrix_to_df `--> `nowcast_matrix_to_df` with same inputs and outputs

kaitejohnson:
It is the former -- we chose to go hyper modular so it is literally just truncating them. The next step `generate_triangles()` takes the list of truncated matrices and converts them to our definition of triangles (replaces bottom right with NA). 

The reason for separating these steps (other than enforcing modularity) is that we use the truncated matrices list as an input for the estimate of the dispersion function since they represent what was observed as of that reference time.

seabbs:
> - we chose to go hyper modular so it is literally just truncating them. 

is this a truncation action then? It sounds like a filter tbh. Anyway that is completely off topic

kaitejohnson:
> > * we chose to go hyper modular so it is literally just truncating them.
> 
> is this a truncation action then? It sounds like a filter tbh. Anyway that is completely off topic

It removes the last row.

Issue #90, opened 2025-04-15
Edit `validate_triangle()` to not require n_rows > n_col


kaitejohnson:
**Goal**
It just needs to be `n_rows > max_delay + 1` I believe.

kaitejohnson:
No longer relevant after #99

PR #92, opened 2025-04-15
Issue 89 & 90: fix validation for triangle inputs

### LINKED ISSUE #89: Throw error if `n` in `get_delay_estimate()` <= `max_delay`
**Bug**
For the method to work, the number of historical observations to be used must be greater than the `max_delay` (number of columns used in the reporting triangle). If this is not the case, the delay_pmf will be all NAs -- we want the package to throw an informative error message when:
`get_delay_estimate(reporting_triangle, max_delay = max_delay, n <= max_delay)`

kaitejohnson:
## Description

This PR closes #89 and #90. Both of these bugs in behaviour were caught while implementing some of the paper analysis and writing it.  What it does:
- adds a function called `.is_reporting_triangle()` which checks whether what is passed in to `validate_triangle()` is of the form of a reporting triangle, with NAs in the bottom right and non-NAs in the top left
- if this is the case, `validate_triangle()` will require that the following criteria are met:
  - `nrow(triangle) >= ncol(triangle)`
  - `n>= max_delay + 1`

Previously, we had required `nrow(triangle) >= ncol(triangle) + 1` which is not necessary
However, we hadn't explicitly checked that `n>= max_delay + 1` which is necessary because `max_delay+1` is the number of columns used. 

There might be an altogether simpler set of checks that I am missing here, or an alternative one we might need to check... also addressed the gaps in code coverage but it didn't update. 

## Checklist


kaitejohnson:
@seabbs I think this might conflict with parts of #99 -- specifically the `.is_reporting_triangle()` function created here, though I believe you created something similar. 

I think all we then need from this PR are these two changes to `.validate_triangle()`:
- `nrow(triangle) >= ncol(triangle)`
- `n>= max_delay + 1`
However, both can be violated with the ragged reporting. Do we want to remove any of these checks then? If so, I will close this and make sure that #99 got rid of the requirement that `nrow(triangle) >= ncol(triangle) + 1`

PR #94, opened 2025-04-17
Issue 93: Test out handling non-symmetric reporting triangle

### LINKED ISSUE #93: Implement non-symmetric report and reference date indexing point nowcast method
**Goal**
Initially, we decided to limit the method to require that users pass in a reporting triangle with the same indexing for both reference and report times, requiring that both be completely filled in and symmetric. We would like to relax this assumption so that we can estimate a reporting delay and generate a point nowcast when there are either:
- only reference dates one time per week, with daily report dates ( a "short" reporting triangle)
- only report dates one time per week with daily reference dates (a "long" reporting triangle)

In particular, the first case we will need to implement the method that works around the lack of explicit handling of weekday effects in the method, by allowing us to filter to reference dates of a particular weekday. 

We will start by testing out an implementation in a toy case, and then write up the maths and test out in the paper analysis code.

kaitejohnson:
## Description

This PR closes #93. WIP.  To test out a reprex, run the `scratch.R` file. 

The approach I have taken here is likely a bit programming intensive, but essentially the idea is that if you pass in a short or long triangle: e.g.:

```r
short_triangle <- matrix(
  c(
    1, 2, 1, 2, 2, 3, 2, 4,
    1, 2, 1, 2, 2, 3, 2, 4,
    1, 2, 1, 2, 2, 3, NA, NA,
    1, 2, 1, 2, NA, NA, NA, NA,
    1, 2, NA, NA, NA, NA, NA, NA
  ),
  nrow = 5,
  byrow = TRUE
),

long_triangle <- matrix(
  c(
    1, 2, 3, 4,
    2, 4, 6, 8,
    1, 2, 3, NA,
    2, 4, 6, NA,
    1, 2, NA, NA,
    2, 4, NA, NA,
    1, NA, NA, NA,
    2, NA, NA, NA
  ),
  nrow = 8,
  byrow = TRUE
)
```

The method will detect the "interval" with which either the reference times or delays are of the other, e.g. in this case `int = 2`. Or if we were in filtering reference times for a single weekday `int = 7`. Then, it creates a list of symmetric reporting triangles, fills in the triangles, and binds them back together properly (which would then be used to estimate the pmf). 

I tested out on this simple example, for a long and short triangle where I know the correct answer. I don't think we need the redundancy  in `scratch_long_tri_example.R` as we can just detect whether the triangle is long or short and then transpose the matrix, perform the operation column wise (as in the short triangle example), and then transpose the final result back (I think). But would definitely need to test on edge cases. See bottom of `scratch_long_tri_example.R` for a test of this.

Noting this adds quite a bit of complexity to the `get_delay_estimate()` function. I am likely missing a less programming intensive method of updating the iteration process though. 

Also caveat of this is it requires a constant interval (which we would need to write a check for using the NA patterning). 

**Alternative approaches**
Note, I started with the approach of attempting to update the multiplicative iterative process, by taking the block bottom left and instead of taking the sum of each of the rows, summing across the nth columns in each row, but I realised this would also require the mulitplicative factor to be either a row wise or column wise vector, and it seemed like a lot of detection to have to be happening automatically. I think this is where I might just be missing something obvious. 

## Checklist


seabbs:
Shall we close in favour of #99?

kaitejohnson:
> Shall we close in favour of #99?

Yes

PR #98, opened 2025-04-24
Issue 97: Estimate dispersion by delay and reference time

### LINKED ISSUE #97: Make maths and code consistent and estimate dispersion by delay
[**Goal**](url)
Related to #95 and per f2f discussions we are going to:
- modify the maths in the model definition to reflect the fact that we are applying the uncertainty to elements of the point nowcast matrix (instead of incorrectly, as written, the already aggregated point nowcasted estimates)
- modify the code to estimate a dispersion parameter for each delay across all the reference times and retrospective nowcast dates (instead of summing the elements as it is doing now)

In the RESPINOW code, they are not estimating dispersion by reference time -- they estimate it by horizon-- which is how they get around the fact that there is no way to use this method to estimate a dispersion parameter as a function of reference time for the latest reference time since there are no observations to compare it to. 

This means that for each retrospective nowcast, a horizon of `h` corresponds to a different reference time

kaitejohnson:
## Description

This PR closes #97. It does the following:
- makes our code consistent so that we are estimating the dispersion using individual elements of the reporting matrix (comparing observations at specific delays and reference times to point nowcasts at those delays, reference times, and retrospective nowcast dates), and then estimating dispersion, as a function of delay $d$ across all the reference times ( ( ($s^*-d$) and retrospective nowcast times $s^*$. E.g.
```math
X_{t,>d} | \hat{X}_{t=s^*-d, >d}(s^*) \sim NegBin(\mu = \hat{X}_{t=s^*-d, >d}+ 0.1, \phi = \phi_d)
```
- changes the implementation to apply these dispersion estimates to individual elements in the nowcast matrix columnwise (by delay) 
- updates the math in the model definition to reflect this, see my comment about struggling to communicate that we are stacking retrospective nowcast dates $s^*$ and estimating dispersion across reference times and retrospective nowcast dates, for each delay.

Note that this implementation is a lot less dispersed than the KIT method. I am thinking that this has to do with the difference in implementation between assuming a negative binomial observation model in the matrix elements vs in the predicted aggregated counts yet to be observed, bc in the former, we end up summing negative binomial draws (but I had thought this would be accounted for by higher dispersion estimates in the individual elements of the matrix ).

This implementation:
image
Previous implementation:
image
KIT method 
image

## Checklist

### DIFF
--- a/tests/testthat/test-estimate_dispersion.R
+++ b/tests/testthat/test-estimate_dispersion.R
@@ -88,7 +88,7 @@ test_that("Error conditions are properly handled", {
   # Invalid n values
   expect_error(estimate_dispersion(valid_nowcasts, valid_trunc_rts, n = -1))
   expect_error(estimate_dispersion(valid_nowcasts, valid_trunc_rts, n = 1.5))
-  expect_error(estimate_dispersion(valid_nowcasts, valid_trunc_rts, n = 3))
+  expect_error(estimate_dispersion(valid_nowcasts, valid_trunc_rts, n = 4))
 })
 
 ### Test 4: Edge Cases ---------------------------------------------------------
@@ -109,10 +109,7 @@ test_that("Matrix dimension validation works", {
     test_triangle[1:5, ],
     test_triangle[1:3, ]
   )
-  expect_error(
-    estimate_dispersion(valid_nowcasts, bad_trunc_rts),
-    "Dimensions of the first `n` matrices in `pt_nowcast_mat_list` and"
-  ) # nolint
+  expect_error(estimate_dispersion(valid_nowcasts, bad_trunc_rts)) # nolint
 })
 
 ## Test 6: fit_nb returns NA if nothing passed to it---------------------------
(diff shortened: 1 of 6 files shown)


tests/testthat/test-estimate_dispersion.R
```diff
@@ -17,36 +17,46 @@ nowcast1 <- matrix(
...
+    10, 7, 1,
+    15, 12, 2.3,
     14, 16, 2.5,
-    11, 25.5, 2
+    11, 25.5, 2.1
   ),
   nrow = 4,
   byrow = TRUE
 )
+nowcast3 <- matrix(
```
seabbs, line 36: These tests only make sense if at the time of implementation we understand what the code is doing so I think we need more unit tests for the internals here and a bit more clarity on what we expect the output to be?

seabbs (commented):
I will come back to this but I think the tests need beefing up.

tests/testthat/test-estimate_dispersion.R
kaitejohnson: Can add to these some examples where we'd expect very high or low dispersion as a function of delay?

kaitejohnson:
Going to close this per f2f discussion to estimate dispersion after aggregating across delays

PR #99, opened 2025-04-24
Issue 93: ragged reporting support

### LINKED ISSUE #93: Implement non-symmetric report and reference date indexing point nowcast method

seabbs:
## Description

This draft PR closes #93.

It refactors `get_delay_estimates` and adds support for ragged daily reporting by dropping the assumed relationship between rows and columns and instead using NAs to fix row locations.

It leads to knock on findings: 

- .validate_triangles is overly restricted and doesn't supported ragged daily reporting
- Support is also needed everywhere that reporting is constructed (i.e replace NA blah blah). I think that this can be simpler than the adaptive approach used here i.e with an index.
- anywhere that iterative multiplication support is used this needs to be ported. We can do this using our new internal functions.

## Checklist

### DIFF
--- a/R/get_delay_estimate.R
+++ b/R/get_delay_estimate.R
@@ -41,9 +41,11 @@
 #'   n = 4
 #' )
 #' print(delay_pmf)
-get_delay_estimate <- function(reporting_triangle,
-                               max_delay = ncol(reporting_triangle) - 1,
-                               n = nrow(reporting_triangle)) {
+get_delay_estimate <- function(
+  reporting_triangle,
+  max_delay = ncol(reporting_triangle) - 1,
+  n = nrow(reporting_triangle)
+) {
   # Check that the input reporting triangle is formatted properly.
   .validate_triangle(
     triangle = reporting_triangle,
@@ -54,47 +56,136 @@ get_delay_estimate <- function(reporting_triangle,
   # Produce a warning if the bottom left of a reporting trianglei s all 0s.
   if (isTRUE(.check_zeros_bottom_right(reporting_triangle))) {
     cli_warn(
-      message =
-        c(
-          "All entries in bottom right are 0. Are these true observations ",
-          "with zero reports, or are these unobserved? If the latter, ",
-          "replace with NA using the `replace_bottom_right_with_NA()` ",
-          "function."
-        )
+      message = c(
+        "All entries in bottom right are 0. Are these true observations ",
+        "with zero reports, or are these unobserved? If the latter, ",
+        "replace with NA using the `replace_bottom_right_with_NA()` ",
+        "function."
+      )
     )
   }
-  # Filter the reporting_triangle down to nrow = n_history_delay + 1,
-  # ncol = max_delay #nolint
+
+  # Filter the reporting_triangle down to relevant rows and columns
+  trunc_triangle <- .prepare_triangle(reporting_triangle, max_delay, n)
+
+  # Fill in missing values in the triangle
+  expectation <- .fill_triangle(trunc_triangle)
+
+  # Calculate probability mass function from filled triangle
+  pmf <- .calculate_pmf(expectation)
+
+  return(pmf)
+}
+
+#' Prepare the triangle by truncating and handling negative values
+#'
+#' @param reporting_triangle The original reporting triangle
+#' @param max_delay Maximum delay to consider
+#' @param n Number of reference times to use
+#' @return Prepared reporting triangle
+#' @noRd
+.prepare_triangle <- function(reporting_triangle, max_delay, n) {
   nr0 <- nrow(reporting_triangle)
   trunc_triangle <- reporting_triangle[(nr0 - n + 1):nr0, 1:(max_delay + 1)]
   rep_tri <- .handle_neg_vals(trunc_triangle)
+  return(rep_tri)
+}
+
+#' Fill in missing values in the reporting triangle
+#'
+#' @param rep_tri Prepared reporting triangle
+#' @return Matrix with imputed values for missing entries
+#' @noRd
+.fill_triangle <- function(rep_tri) {
   n_delays <- ncol(rep_tri)
   n_dates <- nrow(rep_tri)
-  mult_factor <- vector(length = max_delay - 1)
   expectation <- rep_tri
+
   # Find the column to start filling in
   start_col <- which(colSums(is.na(rep_tri)) > 0)[1]
 
   # Only fill in reporting triangle if it is incomplete
   if (!is.na(start_col)) {
-    for (co in start_col:(n_delays)) {
-      block_top_left <- rep_tri[1:(n_dates - co + 1), 1:(co - 1), drop = FALSE]
-      block_top <- rep_tri[1:(n_dates - co + 1), co, drop = FALSE]
-      mult_factor[co - 1] <- sum(block_top) / max(sum(block_top_left), 1)
-      block_bottom_left <- expectation[(n_dates - co + 2):n_dates, 1:(co - 1),
-        drop = FALSE
-      ]
-      # We compute the expectation so that we can get the delay estimate
-      expectation[(n_dates - co + 2):n_dates, co] <- mult_factor[co - 1] *
-        rowSums(
-          block_bottom_left
-        )
+    for (co in start_col:n_delays) {
+      # Extract relevant blocks of the triangle
+      block_top_left <- .extract_block_top_left(rep_tri, co, n_dates)
+      block_top <- .extract_block_top(rep_tri, co, n_dates)
+
+      # Calculate multiplication factor
+      mult_factor <- .calculate_mult_factor(block_top, block_top_left)
+
+      # Extract block bottom left
+      block_bottom_left <- .extract_block_bottom_left(expectation, co, n_dates)
+
+      # Compute expectations for bottom right
+      expectation[(n_dates - co + 2):n_dates, co] <- .compute_expectations(
+        mult_factor,
+        block_bottom_left
+      )
     }
   }
 
+  return(expectation)
+}
 
-  # Use the completed reporting matrix to get the point estimate of the
-  # delay distribution
-  pmf <- colSums(expectation) / sum(expectation)
-  return(pmf)
+#' Extract the top left block of the triangle
+#'
+#' @param rep_tri Reporting triangle
+#' @param co Current column
+#' @param n_dates Number of dates
+#' @return Top left block of the triangle
+#' @noRd
+.extract_block_top_left <- function(rep_tri, co, n_dates) {
+  return(rep_tri[1:(n_dates - co + 1), 1:(co - 1), drop = FALSE])
+}
+
+#' Extract the top block of the triangle
+#'
+#' @param rep_tri Reporting triangle
+#' @param co Current column
+#' @param n_dates Number of dates
+#' @return Top block of the triangle
+#' @noRd
+.extract_block_top <- function(rep_tri, co, n_dates) {
+  return(rep_tri[1:(n_dates - co + 1), co, drop = FALSE])
+}
+
+#' Extract the bottom left block of the triangle
+#'
+#' @param expectation Expectation matrix
+#' @param co Current column
+#' @param n_dates Number of dates
+#' @return Bottom left block of the triangle
+#' @noRd
+.extract_block_bottom_left <- function(expectation, co, n_dates) {
+  return(expectation[(n_dates - co + 2):n_dates, 1:(co - 1), drop = FALSE])
+}
+
+#' Calculate multiplication factor
+#'
+#' @param block_top Top block
+#' @param block_top_left Top left block
+#' @return Multiplication factor
+#' @noRd
+.calculate_mult_factor <- function(block_top, block_top_left) {
+  sum(block_top) / max(sum(block_top_left), 1)
+}
+
+#' Compute expectations for the bottom right part
+#'
+#' @param mult_factor Multiplication factor
+#' @param block_bottom_left Bottom left block
+#' @return Vector of expectations
+#' @noRd
+.compute_expectations <- function(mult_factor, block_bottom_left) {
+  mult_factor * rowSums(block_bottom_left)
+}
+
+#' Calculate the probability mass function from the filled triangle
+#'
+#' @param expectation Filled triangle matrix
+#' @return Probability mass function
+#' @noRd
+.calculate_pmf <- function(expectation) {
+  colSums(expectation) / sum(expectation)
 }
--- a/R/get_delay_estimate.R
+++ b/R/get_delay_estimate.R
@@ -168,7 +168,7 @@ get_delay_estimate <- function(
 #' @return Multiplication factor
 #' @noRd
 .calculate_mult_factor <- function(block_top, block_top_left) {
-  sum(block_top) / max(sum(block_top_left), 1)
+  return(sum(block_top) / max(sum(block_top_left), 1))
 }
 
 #' Compute expectations for the bottom right part
@@ -178,7 +178,7 @@ get_delay_estimate <- function(
 #' @return Vector of expectations
 #' @noRd
 .compute_expectations <- function(mult_factor, block_bottom_left) {
-  mult_factor * rowSums(block_bottom_left)
+  return(mult_factor * rowSums(block_bottom_left))
 }
 
 #' Calculate the probability mass function from the filled triangle
@@ -187,5 +187,5 @@ get_delay_estimate <- function(
 #' @return Probability mass function
 #' @noRd
 .calculate_pmf <- function(expectation) {
-  colSums(expectation) / sum(expectation)
+  return(colSums(expectation) / sum(expectation))
 }
--- a/R/get_delay_estimate.R
+++ b/R/get_delay_estimate.R
@@ -107,18 +107,24 @@ get_delay_estimate <- function(
   # Only fill in reporting triangle if it is incomplete
   if (!is.na(start_col)) {
     for (co in start_col:n_delays) {
+      start_row <- which(is.na(rep_tri[, co]))[1]
       # Extract relevant blocks of the triangle
-      block_top_left <- .extract_block_top_left(rep_tri, co, n_dates)
-      block_top <- .extract_block_top(rep_tri, co, n_dates)
+      block_top_left <- .extract_block_top_left(rep_tri, co, n_dates, start_row)
+      block_top <- .extract_block_top(rep_tri, co, n_dates, start_row)
 
       # Calculate multiplication factor
       mult_factor <- .calculate_mult_factor(block_top, block_top_left)
 
       # Extract block bottom left
-      block_bottom_left <- .extract_block_bottom_left(expectation, co, n_dates)
+      block_bottom_left <- .extract_block_bottom_left(
+        expectation,
+        co,
+        n_dates,
+        start_row
+      )
 
       # Compute expectations for bottom right
-      expectation[(n_dates - co + 2):n_dates, co] <- .compute_expectations(
+      expectation[start_row:n_dates, co] <- .compute_expectations(
         mult_factor,
         block_bottom_left
       )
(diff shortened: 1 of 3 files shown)


R/get_delay_estimate.R
```diff
@@ -11,6 +11,8 @@
 #' @param reporting_triangle Matrix of the reporting triangle, with rows
 #'   representing the time points of reference and columns representing the
 #'   delays. Can be a reporting matrix or incomplete reporting matrix.
+#'    Can also be a ragged reporting triangle, where multiple columns are
+#'    reported for the same row. (e.g. weekly reporting of daily data).
```
kaitejohnson, line 15: Wouldn't weekly reporting of daily data be a long ragged triangle (you have N reference times, and N/7 reporting times)?

R/get_delay_estimate.R
```diff
@@ -54,47 +58,141 @@ get_delay_estimate <- function(reporting_triangle,
...
   n_delays <- ncol(rep_tri)
   n_dates <- nrow(rep_tri)
-  mult_factor <- vector(length = max_delay - 1)
   expectation <- rep_tri
+
   # Find the column to start filling in
   start_col <- which(colSums(is.na(rep_tri)) > 0)[1]
 
   # Only fill in reporting triangle if it is incomplete
   if (!is.na(start_col)) {
```
kaitejohnson, line 110: If you pass in a complete reporting matrix, won't `start_col` be NA, and this whole step of filling in the triangle should be skipped? That is why I had the outer `if(!is.na(start_col))`

R/get_delay_estimate.R
```diff
@@ -54,47 +58,141 @@ get_delay_estimate <- function(reporting_triangle,
...
-      block_bottom_left <- expectation[(n_dates - co + 2):n_dates, 1:(co - 1),
-        drop = FALSE
-      ]
-      # We compute the expectation so that we can get the delay estimate
-      expectation[(n_dates - co + 2):n_dates, co] <- mult_factor[co - 1] *
-        rowSums(
-          block_bottom_left
-        )
+    for (co in start_col:n_delays) {
+      start_row <- which(is.na(rep_tri[, co]))[1]
```
kaitejohnson, line 112: This is nice -- confirming that as written this actually wouldn't require the NA pattern to be regular, because for each column, we identify the earliest row to be filled in iteratively?

tests/testthat/test-get_delay_estimate.R
```diff
@@ -49,11 +42,26 @@ test_that("get_delay_estimate handles partially complete reporting triangles", {
   # Test 8: Test that a partial triangle works correctly
   partial_triangle <- matrix(
     c(
-      80, 50, 25, 10,
-      100, 50, 30, 20,
-      90, 45, 25, 20,
-      80, 40, 15, NA,
-      70, 30, NA, NA
+      80,
```
kaitejohnson, line 45: Any particular reason for this syntax? I find it much easier to read if it looks like the matrix you're making

tests/testthat/test-get_delay_estimate.R
```diff
@@ -122,3 +175,75 @@ test_that("get_delay_estimate handles partially complete reporting triangles", {
...
+
+test_that("get_delay_estimate works with weekly reporting of daily data", {
+  sim_delay_pmf <- c(0.1, 0.2, 0.3, 0.1, 0.1, 0.1, 0.1)
+
+  counts <- c(100, 20, 30, 40, 50, 60, 70)
+
+  complete_triangle <- lapply(counts, function(x) x * sim_delay_pmf)
+  complete_triangle <- do.call(rbind, complete_triangle)
+
+  reporting_triangle <- matrix(
```
kaitejohnson, line 187: Isn't this a 7x7? Wouldn't weekly reporting of daily data be a 7 x 1, 14 x 2 etc? Also think we should test out the short triangle which would have a triangle of  n/7 x n ...

kaitejohnson (commented):
Just some preliminary thoughts/questions.

I think a short triangle example that we'd want to include, which should be treated the same as our weekday filtering is weekly data with daily reporting (not that this is common but this would be where you would have counts only assumed to a reference time on a saturday but allow for adjustments each day)

kaitejohnson:
>  Support is also needed everywhere that reporting is constructed (i.e replace NA blah blah). I think that this can be simpler than the adaptive approach used here i.e with an index.

I think we have to tell this function the "pattern" of the ragged triangle, because if its completely filled in theres no way of knowing the spacing? Unless it assumes that the last column will be entirely NAs and then has to work out the interval from there? (and then we have to filter the data prior to passing it to `replace_bottom_right_w_NA()` )

R/get_delay_estimate.R
seabbs: No I don't think so. This is the short triangle example?

R/get_delay_estimate.R
seabbs: Ah I was thrown off by:  reporting_triangle as the input name can add back in.

R/get_delay_estimate.R
seabbs: yes

tests/testthat/test-get_delay_estimate.R
seabbs: its the air formatter we said we should use. I don't feel strongly about it

R/get_delay_estimate.R
seabbs: wait there is a is.na check here so this case is covered?

R/get_delay_estimate.R
kaitejohnson: Am I losing my mind -- daily data corresponds to the reference times, delays correspond to the report dates. There would be 1/7 of the report dates for each reference time which should make for a long triangle (N x N/7)?

Whereas weekly data with potentially daily updates would be a short triangle (data only assigned to one weekday reference time means N/7 x N matrix)

tests/testthat/test-get_delay_estimate.R
kaitejohnson: Looks nice otherwise, I just prefer these to be consolidated

R/get_delay_estimate.R
seabbs: Daily data with weekly reporting requires fitting 7 nowcasts so at the point estimate the delay then matrix will be just weekly reference times and daily report times

seabbs: Weekly reporting with daily data is easy as its just the normal reporting triangles but with meaningful zeros

R/generate_triangles.R
```diff
@@ -25,27 +26,65 @@
...
+#'     82, 42, NA, NA,
+#'     70, NA, NA, NA
+#'   ),
+#'   nrow = 7,
+#'   byrow = TRUE
+#' )
+#'
+#' retro_rt <- generate_triangle(triangle)
+#' retro_rt
+generate_triangle <- function(reporting_triangle, structure = 1) {
```
seabbs, line 83: I suggest we make .replace.... internal and suggest this for users

seabbs:
Apologies this has widened quite a bit. 

I have addressed the original issue and the knock-on features required everywhere up to`estimate_dispersion` (as that is currently being reword, it will need to make use of the structure handling (potentially internally and maybe as an arg).

The solution implemented is as above drop indexing everywhere for rows based on columns and instead use NAs as the identifier. This means we can automate much of the row indexing which greatly simplifies things. The downside here is that we now need these NAs to be there. I think this isn't an issue as for all real-world use cases where this is the case it should only ever see data without NAs. 

To support generating triangles I have added a `structure` argument which takes either a numeric or a vector if a single value it uses this to implement reporting on that cadence, the vector gives a flexible reporting structure. In a new issue I will add support for a matrix of "example reporting" which will internally be mapped to the current vector support (this will be the easiest user facing interface as it basically says I want past reporting in the cadence of my current reporting). 

The ragged support I think makes the trunc_reporting_triangle data construct outdated and so I have removed it. In making this update I also removed the verbose referring to all functions by name (even when the input and arg had the same name). I think this nicely streamlines the docs and is more in line with both usual and best practice. 

In reviewing/updating the tests I have updated from the mix of styles/layouts being used (I usually give my llm a sample to work with) and also improved tests strategically to try and make sure the new features work as intended. this is mostly aimed at expanding testing to show correctness.

When updating the vignette I had trouble with the `epinowcast` reporting triangle due to the zeros present. I decided it was a better bet to instead start the vignette from the long form data.frame in epinowcast. This has the advantage that it doesn't require the 0 handling epinowcast uses internally, it is in a form people might have their data, and it doesn't add a layer of complexity to think about (i.e why would a NA be shown as a 0). 

More generally package support for NAs as zeros has gone down (I removed the check as it was tricky to refactor). I don't think this is an issue as I don't think it is likely people will encode their data like this (as it isn't natural in a data.frame format to do this). if we want the loose indexing of rows then I think this is the only viable choice.

tests/testthat/test-get_delay_estimate.R
seabbs: I have turned off AIR

kaitejohnson:
> If you mean daily reference dates with only weekly reporting I think that that is best supported using the long form of that matrix (i.e there are report dates with 0 reports) and you take the same approach to splitting the nowcast up by day to get matching periods. If you mean you want to do the latter but fit all the data at once by pretending the periodicity isn't there then I think we could consider that as a feature but I think it needs its own issue and is low priority (as not needed in the paper and I think has a few nuances to it).

This is what I was referring to. I guess my point was that it seems a bit odd to offer jagged support for "short" reporting triangles and not jagged support for "long" reporting triangles (which is what this would be). I agree it is low priority as 1 I don't think its super common and 2 I think as long as we say somewhere that we expect 0s for the report dates for daily reference and weekly reporting such that you pass in a daily-daily matrix whereas we will allow once a week reference dates (or weekly) with daily reporting (even though this is a weekly-daily matrix and violates the general "rule" of keeping the granularity of reference and report dates the same).

R/get_delay_estimate.R
```diff
@@ -40,61 +42,144 @@
...
+  return(expectation[start_row:n_dates, 1:(co - 1), drop = FALSE])
+}
+
+#' Calculate multiplication factor
+#'
+#' @param block_top Top block
+#' @param block_top_left Top left block
+#' @return Multiplication factor
+#' @noRd
+.calculate_mult_factor <- function(block_top, block_top_left) {
```
kaitejohnson, line 164: I know having all of these small functions is maximally modular, but I do have the slightest preference, for the purpose of code readability, for this, 
`.compute_expectations()` and `.calculate_pmf()` to just be in the `.fill_triangle()` function

R/utils.R
```diff
@@ -1,91 +1,137 @@
...
-  # Create a logical mask for valid NA positions
   n_rows <- nrow(mat)
-  mask <- matrix(FALSE, nrow = n_rows, ncol = ncol(mat))
+  n_cols <- ncol(mat)
 
-  for (i in seq_len(n_rows)) {
-    cutoff <- n_rows - i + 1
-    if (cutoff < ncol(mat)) {
-      mask[i, (cutoff + 1):ncol(mat)] <- TRUE
+  for (i in 1:n_rows) {
```
kaitejohnson, line 105: This is nice

R/utils.R
```diff
@@ -1,91 +1,137 @@
...
-  mask <- matrix(FALSE, nrow = n_rows, ncol = ncol(mat))
-
-  for (i in seq_len(n_rows)) {
-    cutoff <- n_rows - i + 1
-    if (cutoff < ncol(mat)) {
-      mask[i, (cutoff + 1):ncol(mat)] <- TRUE
+    if (length(na_indices) > 0) {
+      min_na_idx <- min(na_indices)
+      if (!all(is.na(col_data[min_na_idx:n_rows]))) {
+        return(FALSE)
```
kaitejohnson, line 127: Looks like we should add tests to check column consistency

R/utils.R
```diff
@@ -1,91 +1,137 @@
...
+      }
+      # Check that all entries above the first NA are not NA
+      if (min_na_idx > 1 && anyNA(col_data[1:(min_na_idx - 1)])) {
```
kaitejohnson, line 130: And same here to check for NAs above first NA in the column

R/validate.R
```diff
@@ -19,10 +20,25 @@
...
   assert_matrix(triangle, all.missing = FALSE)
 
-  if (nrow(triangle) <= ncol(triangle)) {
+  # Check if the triangle has a valid structure
+  # Ensure each column has at least one non-NA value
+  if (any(colSums(!is.na(triangle)) == 0)) {
     cli_abort(
-      message =
-        "Number of observations(rows) must be greater than the number of columns." # nolint
+      message = c(
```
kaitejohnson, line 27: This is a nice way of phrasing it that I never thought about

kaitejohnson: We should add tests that this works (though I think this is on your list of new issues already)

tests/testthat/test-apply_delay.R
```diff
@@ -206,3 +207,45 @@ test_that("apply_delay function works the same as the more verbose for loop", {
 
   expect_identical(result, expectation2)
 })
+
+test_that("apply_delay works with ragged reporting triangles", {
+  ragged_triangle <- matrix(
```
kaitejohnson, line 212: I am missing something, shouldn't this be wide? As written its not testing the raggedness

R/get_delay_estimate.R
kaitejohnson: Yes, sorry I think I read this as having got deleted before

tests/testthat/test-get_delay_estimate.R
```diff
@@ -1,124 +1,161 @@
...
+    # Get delay estimate
+    delay_pmf <- get_delay_estimate(
+      reporting_triangle = reporting_triangle
+    )
+    # Test that the function returns the expected PMF
+    expect_identical(delay_pmf, sim_delay_pmf)
+  }
+)
+
+test_that("get_delay_estimate handles complete reporting triangles", {
```
kaitejohnson, line 135: By complete do you mean symmetric?

tests/testthat/test-get_delay_estimate.R
```diff
@@ -1,124 +1,161 @@
...
   )
 
-  expect_warning(get_delay_estimate(
-    reporting_triangle = zero_triangle2,
-    max_delay = 3,
-    n = 5
-  ))
+  expect_is(delay_pmf, "numeric")
+  expect_equal(sum(delay_pmf), 1, tolerance = 1e-6)
+  expect_equal(delay_pmf, partial_pmf, tolerance = 1e-6)
```
kaitejohnson, line 160: Should we be adding tests for `.fill_triangle()` or are these tests of the wrapper fine?

kaitejohnson: I would think it would be good to test that one in particular.

tests/testthat/test-replace_lower_right_with_NA.R
```diff
@@ -1,100 +1,222 @@
...
+  expected_ragged <- matrix(
+    c(
+      1, 3, 5, 7, 9,
+      4, 7, 8, 0, 12,
+      9, 10, 0, 0, 15,
+      3, 0, 0, 0, NA,
+      6, 2, NA, NA, NA
+    ),
+    nrow = 5,
+    byrow = TRUE
```
kaitejohnson, line 167:
```suggestion
  expected_ragged <- matrix(
    c(
      1, 3, 5, 7, 9, 10,
      4, 7, 8, 0, 12, 7,
      9, 10, 0, 0, 15, 10,
      3, 0, 0, 0, NA, NA,
      6, 2, NA, NA, NA, NA
    ),
    nrow = 5,
    byrow = TRUE
```

kaitejohnson: I know this will work but slight preference that this example has the same structure throughout (rather than one NA on the second to last row

kaitejohnson (commented):
This looks mostly good, just a few questions/suggestions on testing and modularising components of the delay estimate function.

R/generate_triangles.R
```diff
@@ -25,27 +26,65 @@
...
+#'     70, 40, 20, 5,
+#'     80, 50, 10, 10,
+#'     100, 40, 31, 20,
+#'     95, 45, 21, NA,
+#'     82, 42, NA, NA,
+#'     70, NA, NA, NA
+#'   ),
+#'   nrow = 7,
+#'   byrow = TRUE
+#' )
```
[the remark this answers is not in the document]
seabbs: The data you have replaced here is important to test on/show as it is what you get when you truncate partially reported data. Happy to change the argument in both generate_triangle and generate_triangles

R/generate_pt_nowcast_mat_list.R
```diff
@@ -116,7 +110,7 @@ generate_pt_nowcast_mat_list <- function(reporting_triangle_list,
 #' point_nowcast_matrix <- generate_pt_nowcast_mat(
 #'   reporting_triangle = triangle
 #' )
-#' print(point_nowcast_matrix)
+#' point_nowcast_matrix
```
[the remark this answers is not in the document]
seabbs: I don't think you need the explicit print

R/generate_triangles.R
seabbs: I am suggesting we remove .replace entirely and use this interface but this was a significant code change we didn't explicitly need here. I think it should be addressed in a new issue

R/get_delay_estimate.R
seabbs: It is also about testing and reuse (i.e we had several copy and pasted bits of code). If you only have a slight preference I would prefer to keep it as it is.

tests/testthat/test-get_delay_estimate.R
seabbs: I mean complete in the sense you have used in the docs I am not sure what you mean by symmetric?

tests/testthat/test-get_delay_estimate.R
seabbs: I think we can though I don't think we need to address it in this PR as the functionality should be covered that is being added here. I think the tests as a whole need some work.

tests/testthat/test-replace_lower_right_with_NA.R
seabbs: I'm not sure I understand sorry? Don't we want to test different structures and sizes?

R/utils.R
seabbs: Same as above covered in to do list

R/get_delay_estimate.R
seabbs: I'm leaving this as is for the reasons given and closing but happy to discuss further in an issue.

R/generate_triangles.R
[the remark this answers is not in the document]
seabbs: As I noted in the main text I dont think the notion of a reporting matrix, incomplete reporting matrix and reporting triangle really makes sense now we have ragged triangle support. I think we need only complete and incomplete i.e triangle and matrix? Things that can only support complete would have a matrix input and things that can also support a triangle would take a triangle. 

All of this aside as this function is just following generate_triangles I think we should address any renaming in a new issue.

R/get_delay_estimate.R
seabbs: See #110 I am not sure what other use cases you had in mind for this. I think given I don't fully understand the use case anything else might be better addressed in a new issue.

seabbs:
> This is what I was referring to. I guess my point was that it seems a bit odd to offer jagged support for "short" reporting triangles and not jagged support for "long" reporting triangles (which is what this would be). I agree it is low priority as 1 I don't think its super common and 2

I have opened a new issue for this as I think it is documentation vs a feature. I don't really understand the call for fitting a "long" format as I don't think the delay you would get would make sense.

seabbs:
I have addressed your review points, opened new linked issues where it seemed needed, and coverage the test gaps mentioned in the last update.

R/generate_triangles.R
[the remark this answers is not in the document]
kaitejohnson: Apologies, I think I got confused when reviewing this, and agree in rejecting the suggestion. 

I think we can just make a separate issue for sorting this out as is described in #108, where the input should not be named a reporting triangle

R/utils.R
kaitejohnson: I think it was just that this code hadn't been tested but I see that these tests have been added.

kaitejohnson: (and perhaps were there and I missed them)

tests/testthat/test-get_delay_estimate.R
kaitejohnson: But then you would want? Because the test should make sure that we can get a delay_pmf from a complete reporting matrix. I think I was suggesting "symmetric" meaning 1:1 reference:report because I thought thats what you meant to be testing here. 
```{r}
  delay_pmf <- get_delay_estimate(
    reporting_triangle = partial_complete,
    max_delay = 3,
    n = 4
  )
  ```

File: tests/testthat/test-replace_lower_right_with_NA.R (hunk 15 above)
Comment (kaitejohnson) [reply tests/testthat/test-replace_lower_right_with_NA.R:167]: Yeah I guess its fine -- I know you want to test the random jaggedness but I also think the example makes more sense and is more realistic when the jaggedness is consistent (e.g. `structure = c(2, 2, 2)`)

Comment (kaitejohnson) [reply tests/testthat/test-replace_lower_right_with_NA.R:167]: But I don't feel strongly so probably fine as is.

**Comment by kaitejohnson** [comment]
I think this all looks good and seems reasonable --  thank you for adding the new issues. This is the only remaining thing I am still not understanding https://github.com/epinowcast/baselinenowcast/pull/99#discussion_r2075541838

File: tests/testthat/test-get_delay_estimate.R (hunk 13 above)
Comment (seabbs) [inline tests/testthat/test-get_delay_estimate.R:135]:
```suggestion
test_that("get_delay_estimate handles diagonal reporting triangles", {
```

seabbs: I see what you mean. The complete matrix step is being tested above. I have clarified that this is a diagonal reporting triangle. Happy to have an issue to discuss nomenclature further perhaps as part of the other suggested renaming.

seabbs:
I've addressed so repinged for a review.

kaitejohnson (approved):
This looks good to me. Seeing the diagonal reporting triangle language (which I like) and we should add to the nomenclature/I should also review and make sure everything is actually consistent... 

But for a separate PR. This looks good, noting it won't work for the uncertainty estimate but we can address that after #104 since that will refactor dispersion estimate anyway.

PR #104, opened 2025-04-25
Issue 103: Implement dispersion by horizon applied to aggregate nowcasts

### LINKED ISSUE #103: Test out implementing dispersion following KIT method
This should very closely approximate the code in the RESPINOW Hub. The distinction between this and what we had previously in main is that this should be estimated and applied after aggregating across delays, and only to the elements in the point nowcast matrix that were not already observed.

kaitejohnson:
## Description

This PR closes #103. At a high-level, it changes the user workflow so that instead of estimating dispersion at the level of individual elements of the matrix, it estimates dispersion aggregated across reporting delays by reference time, so the user only accesses probabilistic nowcasts as vectors. How it does this:
- creates new function `extract_predictions()` which simply takes a matrix and the reporting triangle and extracts only the elements of the first matrix that aren't observed (are NA) in the reporting triangle
- creates new functions `get_nowcast_pred_draw(s)()` which ingests a point nowcast prediction matrix/list of matrices and dispersion parameters and generates a draw/`n` draws of the negative binomial observation model. This will replace `add_uncertainty()`
- modifies `estimate_dispersion()` to now estimate dispersion by first aggregating across reporting delays (`rowSums`) and comparing observed counts at each reference time to the nowcasted counts at each reference time for all delays that needed to be estimated
- deletes the following functions: 
   - `add_uncertainty()`
   - `get_nowcast_matrix_to_df()`
   - `aggregate_df_by_ref_time()`
   - `get_nowcast_mat_draw()`
- Updates the `model_definition.Rmd` vignette to reflect the changes to the dispersion estimate implemented.\
- Updates the `baselinenowcast.Rmd` vignette to reflect this new function flow. 

Remaining question:
I think this should be for a separate PR, but in the final step we combine using dplyr the long dataframe format of the original reporting triangle and the predicted nowcast draws. Do we want to also provide a helper function to do this final step? Otherwise, the last function only returns the predicted nowcast vectors.

Noting that this implementation still leads to underdispersed nowcasts compared to the KIT method as implemented, and this will be addressed in #107 where we will allow an additional mapping function (beyond just `rowSums`, which is what is currently being used) to be applied to the nowcast and reporting matrices to generate the "target", of which the dispersion will be estimated based upon. 
image
image

## Checklist


seabbs:
I thought we agreed not to implement this into the package - at least for now?

kaitejohnson:
> I thought we agreed not to implement this into the package - at least for now?

Per f2f discussion, we agreed to implement this into the package (aggregating across delays before estimating dispersion). I've made a separate issue to optionally pass in a function that maps to another transformation to create the vector by reference date (and will do that separately). For now, this PR will aggregate across delays to estimate dispersion on a vector of reference times.

seabbs (approved):

PR #119, opened 2025-05-07
Fix `generate_pt_nowcast_mat/_list`

### LINKED ISSUE #113: Make _list versions of functions have the same arguments aside from being a list helper
https://github.com/epinowcast/baselinenowcast/blob/9c8ae5f2c2eb1bbeeb8751a3f853a1b660bdf1fd/R/generate_pt_nowcast_mat_list.R#L54

i.e here it is missing the delay_pmf option

kaitejohnson:
## Description

This PR closes #113. We move the error checking to the lower level function, now the entire `_list` function is just:

The entire list function in this case is now: 
```r
  pt_nowcast_mat_list <- lapply(reporting_triangle_list,
    generate_pt_nowcast_mat,
    n = n,
    delay_pmf = delay_pmf
  )
```

I didn't catch any others where the arguments in lower level function are missing in the `_list` version. 
I think #113 suggests as a follow up we could get rid of the list function.

## Checklist


kaitejohnson:
Yeah I think it's a bit redundant but theres something nice to having it for the sake of a clean function workflow. I think we should fix this particular issue (missing input) and punt to a new issue for removing the `_list` functions (which I think you already made).

seabbs:
I don't think there is an issue for that. The other issue is about safe iterators.

R/generate_pt_nowcast_mat_list.R
```diff
@@ -47,33 +49,12 @@ generate_pt_nowcast_mat_list <- function(reporting_triangle_list,
...
-    )
-  }
-
+                                         ),
+                                         delay_pmf = NULL) {
   pt_nowcast_mat_list <- lapply(reporting_triangle_list,
     generate_pt_nowcast_mat,
-    n = n
+    n = n,
+    delay_pmf = delay_pmf
```
seabbs, line 57: This isn't an argument for the function . If it were to be should it be a list or a single entry. If both should there be tooling to map between these options. max_delay is an argument but isn't passed.

seabbs (commented):
Still some more to do here. Surprised not caught by package checks.

R/generate_pt_nowcast_mat_list.R
kaitejohnson: Missed `max_delay`. `delay_pmf` was an argument. I have just added support for it to be either a single vector or a list. 

I think we will probably want to add a function that can generate this list of `delay_pmf`s, or at the very least show how one would get that with an `lapply ` on `get_delay_estimate()` iterating across a list of retrospective reporting triangles from presumably a different strata than the list being passed here (which is where the delay pmfs will be applied).

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -47,7 +48,27 @@ test_that("Function returns correctly structured output", {
   expect_false(anyNA(result[[2]]))
 })
 
-### Test 2: Default n_history_delay Calculation
+### Function takes in separate delay_pmf
```
seabbs, line 51:
```suggestion
```

This duplicates what is in the test string

tests/testthat/test-generate_pt_nowcast_mat.R
```diff
@@ -82,3 +82,30 @@ test_that("generate_pt_nowcast_mat: Output dimensions match input", {
...
+  ) # nolint
+})
+
+test_that("generate_pt_nowcast_mat can take in another delay PMF", {
+  external_delay_pmf <- c(0.1, 0.1, 0.5, 0.3)
+  result <- generate_pt_nowcast_mat(
+    test_triangle,
+    delay_pmf = external_delay_pmf
+  )
+  expect_equal(result[4, 4] / sum(result[4, ]), 0.3, tol = 1e-3)
```
seabbs, line 110: This seems okay but why not test complete recovery?

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -26,8 +26,9 @@ test_triangle_2 <- matrix(
 )
 
 retro_rts_list <- list(test_triangle_1, test_triangle_2)
+delay_pmf <- c(0.1, 0.1, 0.5, 0.3)
 
-### Test 1: Basic Functionality
+###  Basic Functionality
```
seabbs, line 31:
```suggestion
```

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -47,7 +48,27 @@ test_that("Function returns correctly structured output", {
...
+
+  expect_identical(result1, result2)
+
+  expect_error(generate_pt_nowcast_mat_list(
+    reporting_triangle_list = retro_rts_list,
+    delay_pmf = rep(list(delay_pmf), length(retro_rts_list) - 1)
+  ))
+})
+
+### Default n_history_delay Calculation
```
seabbs, line 71:
```suggestion
```

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -71,26 +92,18 @@ test_that("Default n_history_delay uses minimum rows", {
 
 ### Test 3: Custom n_history_delay
```
seabbs, line 93:
```suggestion
```

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -71,26 +92,18 @@ test_that("Default n_history_delay uses minimum rows", {
...
 test_that("Custom n_history_delay is respected", {
-  result <- generate_pt_nowcast_mat_list(
+  expect_no_error(generate_pt_nowcast_mat_list(
```
seabbs, line 95: shouldn't you test it actually uses this value?

seabbs (changes requested):
Looks good just some remaining gripes about tests that I think we should knock out as we standardised what we are doing on testing.

tests/testthat/test-generate_pt_nowcast_mat_list.R
```diff
@@ -90,12 +87,19 @@ test_that("Default n_history_delay uses minimum rows", {
...
+  )
+  # Compare to estimate using n=5
+  first_triangle <- result[[1]]
+  delay_pmf <- get_delay_estimate(retro_rts_list[[1]],
+    n = 5
+  )
+  exp_first_triangle <- apply_delay(retro_rts_list[[1]], delay_pmf)
+  expect_equal(first_triangle, exp_first_triangle, tol = 0.0001)
+
   expect_error(generate_pt_nowcast_mat_list(
```
seabbs, line 103: why just change one of these? seems safe to remove. Minor so happy to merge as is.

seabbs (approved):
I still see some of these comment titled in this file that would be nice to use and the testing update is a bit inconsistent but I think its fine to merge as is. Perhaps we have an issue to clean up the test structure as its own thing.

tests/testthat/test-generate_pt_nowcast_mat_list.R
kaitejohnson: Sorry I don't think I am following -- this adds a test that makes sure the first triangle is estimated from an`n_history_delay = 5` by comparing to an isolated instance of estimating a delay and applying it to just the first triangle (just making sure it works correctly within the nested function of `generate_pt_nowcast_mat_list()`

kaitejohnson:
> these comment titled in this file

Sorry what do you mean by this?

tests/testthat/test-generate_pt_nowcast_mat_list.R
seabbs: you have other changes that then aren't tested but just checked they don't error. If the default was 5 for some reason this would introduce a test gap.

seabbs:
I resolved the merge issue and fixed the outstanding test format issues.

seabbs (dismissed):

PR #136, opened 2025-05-14
Issue 133: Use defaults in vignette

### LINKED ISSUE #133: Update vignette so that it uses the "default" behavior
We decided in #86 that we would like the default model behaviour to be:
training volume = 3x max_delay
with 50% used for delay estimation and 50% for uncertainty estimation. 

This means we will want to use 60 days of data for delay estimation and 60 for uncertainty in the vignette with a maximum delay of 40 days. 

**Context**
This relates to #106, where it was unclear how to set this from the vignette alone, as we were just using the defaults of the individual functions (which we also might want to change), rather than this decided upon strucutre for the default of the larger `baselinenowcast()` wrapper.

kaitejohnson:
## Description

This PR closes #133. It adjusts the vignette in the following ways: 
- sets nowcast date to August 1, 2021, rather than July to allow for sufficient training data volume 
- changes the `max_delay` from 40 to 30, again to allow for sufficient training data volume, since we will want to use 3x the max delay for total training volume and the data starts beginning of April
- adds explicit calls to `truncate_triangles()` and `generate_pt_nowcast_mat_list()` to specify 1. the number of reference times to use for uncertainty estimation and 2. the number of reference times to use for delay estimation (within the retrospective nowcasts, which we will set as the same as we are using to generate the point nowcast matrix).

I did not adjust the defaults in `get_delay_estimate`, `truncate triangles` or `generate_pt_nowcast_mat_list()`. These are a functions of the `n_history_delay` and `max_delay`, though I think we could set this up, I am just a bit stuck on how it would work in the ragged case (where you can have more columns than rows). 

## Checklist


R/truncate_triangles.R
```diff
@@ -10,9 +10,8 @@
 #'   time points of reference and columns representing the delays.
 #' @param n Integer indicating the number of retrospective
 #'   truncated triangles to be generated, always starting from the most
-#'   recent reference time. Default is to only generate truncated matrices
-#'   that have sufficient number of rows to generate a nowcast from, though
-#'   any number can be specified.
+#'   recent reference time. Default is to generate triangles for 50% of the
```
seabbs, line 13: This doesn"t look like it actually reflects the n setting.

seabbs: Why isn't this also defaulting to nrow?

vignettes/baselinenowcast.Rmd
```diff
@@ -194,7 +198,7 @@ reporting triangle.
 # observed long above.
 observed_long <- data_long[location == "DE"][age_group == "00+"] |> # nolint
   enw_filter_report_dates(latest_date = nowcast_date) |>
-  enw_filter_reference_dates(include_days = n_history_delay - 1)
+  enw_filter_reference_dates(include_days = n_history_delay + n_history_uncertainty - 1) # nolint
```
seabbs, line 201: you compute this at least twice. I suggest you save it as a variable and reuse it

vignettes/baselinenowcast.Rmd
```diff
@@ -270,9 +274,9 @@ The `get_delay_estimate()` function expects the following inputs:
   - `max_delay`: an integer indicating the maximum delay to estimate. This must
   be less than or equal to the number of rows in `triangle` minus 1, since we
   assume `triangle` is indexed at 0.
-  - `n_history_delay`: an integer indicating the number of observations by reference
+  - `n`: an integer indicating the number of observations by reference
   date to use to fit the delay distribution. This must be less than or equal to
-  the number of rows in `triangle`.
+  the number of rows in `triangle`. Here we will use `n_history_delay`.
```
seabbs, line 279: mention somewhere we could have done the delay and applied it for a point nowcast in one step to better link to the list iterator coming in the next section.

seabbs (commented):
I don't think the treatment of the n uncertainty reflects what is written in the paper or text here and I think that was in written in the text doesn't reflect what is happening in the code. See specific points for detailed comments.

seabbs (commented):
I don't think the treatment of the n uncertainty reflects what is written in the paper or text here and I think that was in written in the text doesn't reflect what is happening in the code. See specific points for detailed comments.

R/truncate_triangles.R
kaitejohnson: What do you mean by the n setting? 
It can't be nrow, it has to use at least max_delay + 1 rows to estimate the delay (thats why I based it off of ncol)

vignettes/baselinenowcast.Rmd
```diff
@@ -369,31 +373,34 @@ nowcast).
...
+
+We will repeat this iterative process for `n_history_uncertainty` reference
+times in the reporting triangle, which are treated as retrospective nowcast
+times. We start from the latest and work backwards.
 
 
 ```{r generate-triangles}
-trunc_rep_tri_list <- truncate_triangles(reporting_triangle)
+trunc_rep_tri_list <- truncate_triangles(reporting_triangle,
+  n = n_history_uncertainty
```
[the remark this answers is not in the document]
kaitejohnson: This `n` is the number of retrospective reporting triangles/retrospective nowcasts we should make. It should be 50% of the reference times. Within each one, we will then use `n_history_delay` to estimate the delay pmf.

kaitejohnson:
> I don't think the treatment of the n uncertainty reflects what is written in the paper or text here and I think that was in written in the text doesn't reflect what is happening in the code. See specific points for detailed comments.

My understanding of the paper default is:
training volume = 3x max_delay
n_history_delay = 50% training volume
n_history_uncertainty = 50% training volume

Which means for `max_delay = 30`, `n_history_delay = 45`, `n_history_uncertainty= 45`. 

When we create the truncated triangles, we need to tell it the `n` number of triangles to create, which in this case should be 45. Then in `generate_pt_nowcast_mat_list` , we have to tell it to use `n_history_delay` to estimate the delay from each of those triangles. This entire uncertainty process requires 90 reference times. 

It is not a clean split of whats being used to estimate delay and whats being used to estimate uncertainty, because within the uncertainty we re-estimate the delay.

R/truncate_triangles.R
kaitejohnson: I think I need to fix the description though it shouldnt have changed

seabbs: You say 50 is the default but it isn't the default its some combination of column and row.

nrow(the data frame) would be longer than that so I am not clear why it can't be?

vignettes/baselinenowcast.Rmd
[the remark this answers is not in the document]
seabbs: its the first 50% so the last   50 % is never used?

seabbs:
> **My understanding of the paper default is:**

Inside the uncertainty function is that how it works though? I think there has been some confusion here. I think it works as I described in the issue. 

Regardless of that I don't think what you have implemented in the code correctly reflects this description.

> This entire uncertainty process requires 90 reference times.

Have you checked you are using all the data? It looks like you filter for the last 45 in both

vignettes/baselinenowcast.Rmd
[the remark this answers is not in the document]
kaitejohnson: for the delay estimate: the last 50% is used to estimate the delay

For the uncertainty estimate, `truncate_triangles()` starts at the end of the data and works backwards. This means that at the least recent nowcast time, it will use the first 50% of the whole training data to estimate the delay for that particular retrospective nowcast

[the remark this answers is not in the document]
seabbs: > means that at the least recent nowcast time, it will 

What does this mean? Are you filtering somewhere so that the data that goes into truncate triangles excludes the data used for estimating the delay?

seabbs:
I sent some code snippets over slack

seabbs:
We discussed face to face with the action being a focus on the documentation of these data requirements being separablef

vignettes/baselinenowcast.Rmd
kaitejohnson: Added a sentence

PR #153, opened 2025-05-19
Refactor .calc_expectation and apply_delay: Correct delay index logic, update tests, and clarify NA handling

seabbs:
This PR refactors the .calc_expectation and apply_delay functions and updates their tests to address issues with delay index handling and NA propagation, as discovered by @kaitejohnson and introduced by @seabbs.\n\n**Summary of changes (from diff with main):**\n- Refactored .calc_expectation to take the full delay_pmf and delay_cdf vectors, and to update delay_prob and delay_cdf_prev only when an NA is found in the current column.\n- Updated apply_delay to pass the correct arguments and iterate using the new logic.\n- Improved documentation for both functions to reflect the new argument structure.\n- Updated and expanded tests for .calc_expectation:\n  - Tests now explicitly check that only the expected cell is modified.\n  - Manual logic is used to verify the value of the updated cell.\n  - All other cells are checked to remain unchanged.\n  - Added/updated tests for ragged triangle structures in apply_delay.\n- Minor .gitignore update for .claude/.\n\n**Context and rationale:**\nThis builds on the discovery by @kaitejohnson, who identified that the previous implementation (introduced by @seabbs) could use a CDF and PMF index that was not aligned with the column being updated, leading to overestimation and error propagation, especially in non-standard (ragged) triangle structures. The new approach ensures that updates only occur when an NA is found, and the correct delay parameters are used for the column being updated.\n\nWe retain the while loop approach because it is safer: it ensures we only update when an NA is found, and avoids propagating errors when the structure is not standard. This is particularly important for ragged triangles, where the first NA may not be in the expected column.\n\n**Closes:**\nThis PR closes the issue described in the branch name and addresses the concerns raised in @kaitejohnson’s review.


seabbs:
Closing in favour of a new, better-structured PR that more clearly documents the bug, its origin, and the fix. See follow-up PR for details.

PR #161, opened 2025-05-21
Issue #159: Fix bugs in dispersion estimate

### LINKED ISSUE #159: Bug in dispersion estimate
**Updated**
Below is an example of a perfect reporting triangle and we are still estimating dispersion at horizons 2 and 3 as way too high... 

```r
 set.seed(123)
  delay_pmf <- c(0.4, 0.3, 0.2, 0.05, 0.05)
  partial_counts <- c(80, 100, 180, 80, 140)
  
  # Create a complete triangle based on the known delay PMF
  triangle <- lapply(partial_counts, function(x) x * delay_pmf)
  triangle <- do.call(rbind, triangle)
  triangle <- rbind(triangle, triangle)
  
  pt_nowcast_mat <- generate_pt_nowcast_mat(triangle)
  trunc_rts <- truncate_triangles(triangle)
  retro_rts <- generate_triangles(trunc_rts)
  
  pt_nowcast_list <- generate_pt_nowcast_mat_list(retro_rts_list)
  
  dispersion <- estimate_dispersion(
    pt_nowcast_list,
    trunc_rep_tri_list,
    rt_list
  )
  dispersion 
#> [1] 999.99991  10.30979  20.99460
```

**Problem**

The call to the Boolean indices `indices_nowcast` is incorrectly point to `triangle_observed` when it should be pointing to the last row of the nowcast backwards to the `k-1`th row of the nowcast.
https://github.com/epinowcast/baselinenowcast/blob/f8c723d02d2a0329b07f1cdde032da67b6c8fd54/R/estimate_dispersion.R#L172

This was a bug introduced in https://github.com/epinowcast/baselinenowcast/pull/131/files#r2098844717. Previously, we had been getting the indices nowcasted by generating a triangle from the truncated observation, and then finding the NAs, and after merging in the support for ragged triangles we removed this and use the retrospective reporting triangle itself. I manually dealt with these merge conflicts, and made an error when applying the new logic here. 

**Additional issue**
The observations will have NAs, so this `rowSums()` needs an `na.rm = TRUE`
https://github.com/epinowcast/baselinenowcast/blob/16478abdee2831987b3338facbdeff1291184012/R/estimate_dispersion.R#L183 in addition to the one around `fun_to_aggregate`.

**Proposed solution**

1. Fix the bug by streamlining this with:
```r
obs <- trunc_matr_observed[start_row:end_row, ]
      tri <- triangle_observed[start_row:end_row, ]
      nowcast <- nowcast_i[start_row:end_row, ]
      indices_nowcast <- is.na(tri)
      indices_observed <- !is.na(obs)
```
2. Add tests to test on k = 1 with negative binomial error on the values of the triangle (note this isn't a perfect test, as I am having trouble of thinking of how to add this error by horizon in a forward process)
3. Add tests to test on k > 1, again with error on the values of the triangle

kaitejohnson:
## Description

This PR closes #159. It addresses two issues in the dispersion estimate. 
1. A bug in how we were extracting the booleans to indicate the indices that were nowcasted and observed, which was accidentally using the indices from the retrospective reporting triangle
2. A bug in how we were computing the `rowSums()` across the matrix when aggregating across delays for the observations. Without a `na.rm = TRUE`, this was always returning 0, which is why, when applied to the COVID data as part of the analysis in `baselinenowcast-paper`, the uncertainty was extremely wide. 

It adds tests that attempt to prevent these kinds of errors from being introduced again. These tests: 
1. test that when you provide the dispersion estimator with perfect data, you get back very high size parameter estimates
2. test that when adding noise to the reporting triangle so it does not have a perfect delay pmf across the triangle, we get realistic dispersion estimates. I think this test could be improved, as I ideally wanted to test back approximate parameter recovery (and then struggled to think about how to do this across the reporting triangle). One thing I could implement is a horizon varying delay_pmf, so that the error is higher for say certain horizons... 
3. test that the dispersion estimates work on `k>1`

As a check that this behavior is what we expect, I have used the vignette to compute the dispersion and nowcasts on the 7d rolling sum. This is in a [branch](https://github.com/epinowcast/baselinenowcast/tree/play-with-7d-sum-in-vignette) that I do not plan on merging. 

The result is below, black is the target data, gray are the median, 50th, and 95th quantiles of the 7d sum nowcast. The better test will of course be in the analysis plan, but there seems to be some issues there that I need to investigate further. 

image

## Checklist

## Summary by CodeRabbit

- **Refactor**
  - Improved internal handling of data segments for more consistent and simplified calculations during dispersion estimation.

- **Tests**
  - Enhanced test coverage for dispersion estimation with new scenarios using perfect and simulated data, and updated existing tests for improved reliability.

- **Documentation**
  - Updated example usage to refine the estimation process parameters for dispersion calculation.

### DIFF
--- a/R/estimate_dispersion.R
+++ b/R/estimate_dispersion.R
@@ -168,10 +168,8 @@ estimate_dispersion <- function(
       start_row <- max_t - d - k + 2
       end_row <- max_t - d + 1
       obs <- trunc_matr_observed[start_row:end_row, ]
-      tri <- triangle_observed[start_row:end_row, ]
       nowcast <- nowcast_i[start_row:end_row, ]
-    indices_nowcast <- is.na(triangle_observed[start_row:end_row, ])
-    indices_observed <- !is.na(trunc_matr_observed[start_row:end_row, ])
+      indices_nowcast <- is.na(triangle_observed[start_row:end_row, ])
       indices_observed <- !is.na(obs)
       # Function to aggregate is always applied after the matrix has been
       # summed across delays


kaitejohnson:
Update: 
I don't think its quite equivalent to say that if we generate samples of the initial reporting triangle from a poisson ( negative binomial with a high dispersion parameter), we should get back poisson distributed errors across retrospective nowcast times and horizons. 

Using a dispersion for all reference times of `size = 500` on the individual elements, we still get back when estimating dispersion as a function of horizon, dispersion that varies across the horizon. 

I think if we were aggregating by reference time, we would expect to recover the original dispersion by reference time, but since we are aggregating by horizon, I think we actually should expect to have lower phi values for the recent horizons and higher as they increase, because the recent horizons we have very few counts to estimate from, and that error gets propagated along the reference time, whereas for the later horizons, nearly the entire reference time's entries are known and so you are adding a value for that horizon that is relatively constrained. 

I went through and debugged into the weeds of each component of the matrices in `exp_to_add` and `to_add_already_observed` and they all look reasonable and what we would expect to see -- for low horizons d, we have more differences across retrospective nowcast times than we do for the higher horizons. 

This of course is consistent with what we expect to observe (increasing uncertainty at recent times), but it is somewhat non-intuitive when setting the uncertainty to be constant across the reporting triangle.

seabbs:
I don't think I agree from a theoretical perspective but I don't think it needs to block this PR.

R/estimate_dispersion.R
```diff
@@ -168,20 +168,21 @@ estimate_dispersion <- function(
       start_row <- max_t - d - k + 2
       end_row <- max_t - d + 1
       obs <- trunc_matr_observed[start_row:end_row, ]
+      tri <- triangle_observed[start_row:end_row, ]
       nowcast <- nowcast_i[start_row:end_row, ]
-      indices_nowcast <- is.na(triangle_observed)[start_row:end_row, ]
-      indices_observed <- !is.na(trunc_matr_observed)[start_row:end_row, ] # nolint
+      indices_nowcast <- is.na(tri)
```
seabbs, line 173: do you need this assigment complexity given obs and ti are used once? To me this makes the code harder to read/reason on

seabbs: inparticular because the order you define them is no the order you then use them in

R/estimate_dispersion.R
```diff
@@ -190,10 +191,10 @@ estimate_dispersion <- function(
   # Estimate the dispersion as a function of horizon across retrospective
   # nowcast dates
   disp_params <- vector(length = n_possible_horizons)
-  for (i in seq_len(n_possible_horizons)) {
-    obs_temp <- to_add_already_observed[, i]
-    mu_temp <- exp_to_add[, i] + 0.1
-    disp_params[i] <- .fit_nb(x = obs_temp, mu = mu_temp)
+  for (j in seq_len(n_possible_horizons)) {
```
seabbs, line 194: motivation?

seabbs (dismissed):
LGTM

R/estimate_dispersion.R
kaitejohnson:
```suggestion
    indices_nowcast <- is.na(triangle_observed[start_row:end_row, ])
    indices_observed <- !is.na(trunc_matr_observed[start_row:end_row, ])
```
I could just use this?

R/estimate_dispersion.R
```diff
@@ -168,20 +168,22 @@ estimate_dispersion <- function(
       start_row <- max_t - d - k + 2
       end_row <- max_t - d + 1
       obs <- trunc_matr_observed[start_row:end_row, ]
+      tri <- triangle_observed[start_row:end_row, ]
       nowcast <- nowcast_i[start_row:end_row, ]
-      indices_nowcast <- is.na(triangle_observed)[start_row:end_row, ]
-      indices_observed <- !is.na(trunc_matr_observed)[start_row:end_row, ] # nolint
+    indices_nowcast <- is.na(triangle_observed[start_row:end_row, ])
+    indices_observed <- !is.na(trunc_matr_observed[start_row:end_row, ])
+      indices_observed <- !is.na(obs)
```
seabbs, line 175:
```suggestion
```

R/estimate_dispersion.R
```diff
@@ -168,20 +168,22 @@ estimate_dispersion <- function(
       start_row <- max_t - d - k + 2
       end_row <- max_t - d + 1
       obs <- trunc_matr_observed[start_row:end_row, ]
+      tri <- triangle_observed[start_row:end_row, ]
```
seabbs, line 171:
```suggestion
```

Are you using obs anywhere else?

seabbs: same point about nowcast

kaitejohnson: Yes obs is used below as is nowcast

seabbs: as in more than once?

seabbs (approved):

R/estimate_dispersion.R
kaitejohnson: twice

R/estimate_dispersion.R
kaitejohnson: keeping the shorter version

PR #174, opened 2025-06-17
Issue 173: Structure requirements are too strict

### LINKED ISSUE #173: `generate_triangles()` should be able to make a triangle with a complete set of NAs as last row
**Bug**
We have required that the sum of the `structure` argument in generate triangles adds up to 1 less than the number of columns, however, in order to generate something like this:

Here's a reprex, where the desired result is  in `exp_result`, and we want to use `generate_triangle()` to create a matrix which has 1, then 3 non NA rows aligned on the left side.

```r
 exp_result <- matrix(
    c(
      1, 3, 5, 7, 9,
      4, 5, 9, 4, 3,
      1, 6, 4, NA, NA,
      3, NA, NA, NA, NA
    ),
    nrow = 4, 
    byrow = TRUE
  )
 
trunc_rt <- matrix(
   c(
     1, 3, 5, 7, 9,
     4, 5, 9, 4, 3,
     1, 6, 4, 4, 3,
     3, 8, 4, 6, 1
   ),
   nrow = 4, 
   byrow = TRUE
 )

actual_result <- generate_triangle(trunc_rt, structure = c(1, 2))
#> Error in `generate_triangle()`: ! Sum of structure vector must equal the number of columns in matrix minus 1
```
since `structure` is counting the number of "new" non-NAs in each row, so in this case we want 1 in the last row, 2 new in the next, and that's it. 

As written, `structure = c(1,2)` will throw an error because the sum is not equivalent to the maximum delay cols minus one. 

Here's what happens in one of the tests for structure:
```r
expected_custom <- matrix(
    c(
      1, 3, 5, 7, 9,
      4, 7, 8, 0, 12,
      9, 10, 0, 0, NA,
      3, 0, 0, NA, NA,
      6, NA, NA, NA, NA
    ),
    nrow = 5,
    byrow = TRUE
  )
  result_custom <- generate_triangle(test_matrix, c(1, 2, 1))
  expect_identical(result_custom, expected_custom)

#> result_custom
     [,1] [,2] [,3] [,4] [,5]
[1,]    1    3    5    7    9
[2,]    4    7    8    0   12
[3,]    9   10    0    0   NA
[4,]    3    0    0   NA   NA
[5,]    6   NA   NA   NA   NA
```
But say we wanted something with 2 non NAs and then another 2 non NAs, like this:
```r
result_custom2
     [,1] [,2] [,3] [,4] [,5]
[1,]    1    3    5    7    9
[2,]    4    7    8    0   12
[3,]    9   10    0    0   8
[4,]    3    0    0   NA   NA
[5,]    6   NA   NA   NA   NA
```
We would would think we could just use `structure = c(1, 2, 2)` but this violates the current check and will throw a warning:
```r
result_custom2 <- generate_triangle(test_matrix, c(1, 2, 2))

#> Error in `generate_triangle()`: ! Sum of structure vector must equal the number of columns in matrix minus 1
```

**Proposed solutions**
- Option 1: Remove the error catch that checks that `sum(structure_vec) =ncols-1` and replace with `sum(structure_vec) <= ncols-1` Pro: Easy to implement. Con: less of a check on structure vec
- Option 2: Consider refactoring the structure code. The problem I see when I try to walk through alternative approaches here is that to maximise flexibility and only have the user provide one vector, either on the left side or right side we wouldn't necessarily have a "check" that the structure_vec spans the columns... 

Something else clever I am not thinking of (certainly)

kaitejohnson:
## Description

This PR closes #173.  It does the following:
- modifies an internal check in `generate_triangle()` which previously had required that the sum of `structure_vec` equal one less than the number of columns. It now requires it be less than one less than the number of columns, enabling multiple NAs in the final row of the triangle. 
- added a test that uses `structure` to try to generate a reporting triangle that contains more than one NA in the last row, this will fail currently

## Checklist

## Summary by CodeRabbit

- **Bug Fixes**
	- Resolved an issue where the sum of elements in the `structure` vector was required to equal exactly one less than the number of columns; it now must be less than the number of columns.
- **Documentation**
	- Updated documentation for the `structure` parameter in relevant functions to clarify the new validation rule.
	- Enhanced examples for `combine_obs_with_pred` to demonstrate usage with larger matrices and rolling sum aggregation.
- **Tests**
	- Added tests to ensure correct handling of matrices with trailing `NA` values and validation of positive integer requirements for the `structure` vector.

### DIFF
--- a/R/generate_triangles.R
+++ b/R/generate_triangles.R
@@ -113,7 +113,7 @@ generate_triangle <- function(trunc_rep_tri, structure = 1) {
 
     if (sum(structure_vec) > (cols - 1)) {
       cli_abort(
-        message = "Sum of structure vector must less than or equal to one less than the number of columns in matrix", # nolint
+        message = "Sum of structure vector must less than or equal to one less than the number of columns in matrix" # nolint
       )
     }
   }
--- a/R/generate_triangles.R
+++ b/R/generate_triangles.R
@@ -65,7 +65,7 @@ generate_triangles <- function(trunc_rep_tri_list, structure = 1) {
 #'    not contain NAs.
 #' @param structure Integer or vector specifying the reporting structure.
 #'   If integer, divides columns evenly by that integer (with last possibly
-#'   truncated).  If vector, the sum must not be greater than one less than the
+#'   truncated).  If vector, the sum must not be greater than or equal to the
 #'   number of columns. Default is 1 (standard triangular structure).
 #' @returns A single retrospective reporting triangle matrix with NAs in the
 #'   appropriate positions.
@@ -113,9 +113,12 @@ generate_triangle <- function(trunc_rep_tri, structure = 1) {
     if (any(structure_vec < 0) || !all(is.finite(structure_vec))) {
       cli_abort("All elements of `structure` must be positive integers")
     }
-    if (sum(structure_vec) > (cols - 1)) {
+    if (sum(structure_vec) >= cols) {
       cli_abort(
-        message = "Sum of structure vector must not be greater than one less than the number of columns in matrix" # nolint
+        message = c(
+          "Sum of structure vector must not be greater than or equal",
+          "to the number of columns in matrix"
+        )
       )
     }
   }
--- a/R/generate_triangles.R
+++ b/R/generate_triangles.R
@@ -110,12 +110,12 @@ generate_triangle <- function(trunc_rep_tri, structure = 1) {
   } else {
     # Vector case
     structure_vec <- structure
-    if (any(structure_vec <= 0) || !all(is.finite(structure_vec))) {
+    if (any(structure_vec < 0) || !all(is.finite(structure_vec))) {
       cli_abort("All elements of `structure` must be positive integers")
     }
     if (sum(structure_vec) > (cols - 1)) {
       cli_abort(
-        message = "Sum of structure vector must be less than or equal to one less than the number of columns in matrix" # nolint
+        message = "Sum of structure vector must not be greater than one less than the number of columns in matrix" # nolint
       )
     }
   }
--- a/R/generate_triangles.R
+++ b/R/generate_triangles.R
@@ -65,7 +65,7 @@ generate_triangles <- function(trunc_rep_tri_list, structure = 1) {
 #'    not contain NAs.
 #' @param structure Integer or vector specifying the reporting structure.
 #'   If integer, divides columns evenly by that integer (with last possibly
-#'   truncated).  If vector, must sum to number of columns.
+#'   truncated).  If vector, be less than one less than the number of columns.
 #'   Default is 1 (standard triangular structure).
 #' @returns A single retrospective reporting triangle matrix with NAs in the
 #'   appropriate positions.
@@ -111,11 +111,10 @@ generate_triangle <- function(trunc_rep_tri, structure = 1) {
     # Vector case
     structure_vec <- structure
 
-    if (sum(structure_vec) != (cols - 1)) {
-      cli_abort(c(
-        "Sum of structure vector must equal the number of columns in matrix",
-        " minus 1"
-      ))
+    if (sum(structure_vec) > (cols - 1)) {
+      cli_abort(
+        "Sum of structure vector must less than or equal to one less than the number of columns in matrix", # nolint
+      )
     }
   }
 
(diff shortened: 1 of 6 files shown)


R/generate_triangles.R
```diff
@@ -110,12 +110,13 @@ generate_triangle <- function(trunc_rep_tri, structure = 1) {
   } else {
     # Vector case
     structure_vec <- structure
-
-    if (sum(structure_vec) != (cols - 1)) {
-      cli_abort(c(
-        "Sum of structure vector must equal the number of columns in matrix",
-        " minus 1"
-      ))
+    if (any(structure_vec <= 0) || !all(is.finite(structure_vec))) {
```
seabbs, line 113: why not zero?

seabbs (commented):
Mostly good some niggles

R/generate_triangles.R
kaitejohnson: forgot that 0 is fine

tests/testthat/test-generate_triangle.R
```diff
@@ -199,6 +199,34 @@ test_that("generate_triangle handles custom structure with vector", {
   expect_identical(result_custom, expected_custom)
 })
 
+test_that("generate_triangle can generate something with all NAs at end", {
```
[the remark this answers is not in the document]
kaitejohnson: For the first element of structure?  I don't think I am seeing the unstated structure you see.

R/generate_triangles.R
```diff
@@ -65,7 +65,7 @@ generate_triangles <- function(trunc_rep_tri_list, structure = 1) {
 #'    not contain NAs.
 #' @param structure Integer or vector specifying the reporting structure.
 #'   If integer, divides columns evenly by that integer (with last possibly
-#'   truncated).  If vector, must sum to number of columns.
+#'   truncated).  If vector, be less than one less than the number of columns.
```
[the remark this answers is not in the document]
seabbs: yes

R/generate_triangles.R
seabbs: ?

seabbs (commented):
Not quite there. The test we need is for the unstated last low structure to show it can be something other than 1 NA or 0 NA.

tests/testthat/test-generate_triangle.R
```diff
@@ -199,6 +199,90 @@ test_that("generate_triangle handles custom structure with vector", {
...
+    nrow = 4,
+    byrow = TRUE
+  )
+  actual_result <- generate_triangle(trunc_rt,
+    structure = c(2, 1)
+  )
+  expect_identical(exp_result, actual_result)
+})
+
+test_that("generate_triangle can handle a structure ending with 2 NAs", {
```
kaitejohnson, line 258: @seabbs I am still not following, it would be helpful if you could just suggest the test you want. 

This test is testing the edge case similar to the problem that occurred which is that the structure must allow for more than one NA at the end (here it is 2, I can make it more if we want).

R/generate_triangles.R
```diff
@@ -65,8 +65,8 @@ generate_triangles <- function(trunc_rep_tri_list, structure = 1) {
 #'    not contain NAs.
 #' @param structure Integer or vector specifying the reporting structure.
 #'   If integer, divides columns evenly by that integer (with last possibly
-#'   truncated).  If vector, must sum to number of columns.
-#'   Default is 1 (standard triangular structure).
+#'   truncated).  If vector, the sum must not be greater than one less than the
```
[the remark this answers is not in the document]
kaitejohnson: It must be less than cols, I will change throughout.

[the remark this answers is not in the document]
kaitejohnson: I don't think that is quite as clear than the maximum value is cols-1 but can change as agreed this is quite wordy

R/generate_triangles.R
kaitejohnson: What do you mean done? Are 0s used?

tests/testthat/test-generate_triangle.R
seabbs: The issue here is that in all of these the implied last report has 0 NAs as you are specifying the reporting for line 262. We want to check that the code can handle settings where the line for 261 is not zero or 1 NA but some other number as that is the implied and not explicit structure. That is currently missing from the tests. As this is a sticking point I think its fine to write a new issue to address this. I don't have time to work on this for the next two weeks.

seabbs (approved):

PR #208, opened 2025-07-21
Issue 115: refactor estimate dispersion

### LINKED ISSUE #115: Refactor estimate_dispersion
Related to #112. As currently implemented `estimate_dispersion` does multiple tasks across lists of input including input checks, a simple nowcast wrapper, and finally fitting. It seems sensible for these to be split into single iterated tasks as elsewhere in the code base (potentially as a helper for others or potentially internal only facing). It also seems sensible not to hard code the error model being fit so enable more downstream functionality. 

I propose the following: 

- Write a single low-level function to handle the current point nowcast steps generalising to take 3 inputs vs 2 i.e the data to be nowcast, the point nowcast, and the finally observed data
- Write another function to handle aggregating a point nowcast with input functions to handle different schemes
- Write a standalone function for fitting the output of these that can take a fitting function of the standardised form
- Write a wrapper for the above renamed to be more general i.e `estimate_uncertainty`, add another wrapper i.e estimate_uncertainty_from_triangle for the simple case. Another slightly more complex option here is to make `estimate_uncertainty` and s3 method which would be my preference
- Rewrite add_uncertainty to apply uncertainty from any function vs hardcoding as now.

I see some overlap here with #104

kaitejohnson:
## Description

This PR closes #115. A bit of a doozy. 
Changes made: 
- Written a standalone function `estimate_uncertainty` that ingests the point nowcast matrices, the truncated reporting triangles, and the retrospective reporting triangles, an error model and its arguments, and the aggregator functions for reference times and delays 
- Written a standalone function for fitting observations and predictions `fit_distribution` that is by default passed to estimate uncertainty as the `error_model`, which takes in the observations and predictions and a character of a parametric distribution and returns a vector of uncertainty parameters for each  column (in this case horizon) 
- Written a standaone function for sampling from an observation model given a set of predictions and uncertainty parameters, `sample_distribution`
- made `sample_prediction`, `sample_predictions`, `sample_nowcast`, `sample_nowcasts`, and `combine_obs_with_pred` which all take in the error model used and the aggregator functions and there are arguments (`combine_obs_with_pred` only needs aggregator functions).
- adds tests for ragged triangle and rolling sum aggregator functions

What we haven't done yet:
- tested out different `delay_aggregators` (e.g. what if there is none and you have a vector of uncertainty parameters of a different length for each horizon) 

## Checklist

## Summary by CodeRabbit

* **New Features**
  * Added flexible support for multiple observation models (negative binomial, normal, gamma) for uncertainty estimation and prediction sampling.
  * Introduced new functions for fitting (`fit_distribution`) and sampling (`sample_distribution`) from statistical distributions.
  * Enabled customizable aggregation and error model functions for nowcast uncertainty estimation and prediction generation.
  * Added a new helper function for masked matrix operations.

* **Bug Fixes**
  * Improved input validation and error handling for observation models and aggregation parameters.
  * Enhanced consistency checks and warnings for insufficient data in uncertainty estimation.

* **Documentation**
  * Updated and expanded documentation to reflect new parameters, usage patterns, and supported models.
  * Added documentation for new internal and exported functions, including `.apply_mask`, `fit_distribution`, and `sample_distribution`.
  * Revised existing function docs to describe flexible aggregation and error model parameters.

* **Tests**
  * Added comprehensive unit tests for new and updated functions, covering input validation, output correctness, and statistical properties.
  * Updated existing tests to align with the new flexible interfaces and expanded coverage for multiple observation models.
  * Removed tests for obsolete aggregation validation and replaced with tests for new model support.

* **Refactor**
  * Generalized core logic to support arbitrary error models and aggregation strategies, replacing hardcoded sum and dispersion approaches.
  * Removed obsolete helper functions and related documentation.
  * Simplified input assertions and validation logic.


R/extract_observed_nowcast_components.R
```diff
@@ -0,0 +1,244 @@
...
+    for (d in 1:n_horizons) {
+      start_row <- max_t - d - k + 2
+      end_row <- max_t - d + 1
+      obs <- trunc_matr_observed[start_row:end_row, ]
+      nowcast <- nowcast_i[start_row:end_row, ]
+      indices_nowcast <- is.na(triangle_observed[start_row:end_row, ])
+      indices_observed <- !is.na(obs)
+      # Function to aggregate is always applied after the matrix has been
+      # summed across delays
+      exp_to_add[i, d] <- fun_to_aggregate(
```
kaitejohnson, line 228: `fun_to_aggregate` and `rowSums` are passed in functions

R/estimate_uncertainty.R
```diff
@@ -135,68 +148,46 @@ estimate_uncertainty <- function(
...
-  } # end loop over retrospective nowcast times
 
-  # Estimate the dispersion as a function of horizon across retrospective
-  # nowcast dates
-  disp_params <- vector(length = n_possible_horizons)
-  for (j in seq_len(n_possible_horizons)) {
-    obs_temp <- to_add_already_observed[, j]
-    mu_temp <- exp_to_add[, j] + 0.1
-    disp_params[j] <- .fit_nb(x = obs_temp, mu = mu_temp)
+    # Apply the aggregation
```
kaitejohnson, line 159: Aggregating first to get the matrices where the row wise aggregation is in the latest reference times location (will reduce the matrix rows if k>1)

R/estimate_uncertainty.R
```diff
@@ -135,68 +148,46 @@ estimate_uncertainty <- function(
...
+    aggr_obs <- do.call(aggregator, c(list(trunc_matr_observed), aggregator_args))
+    aggr_nowcast <- do.call(aggregator, c(list( nowcast_i), aggregator_args))
+    aggr_rt_obs <- do.call(aggregator, c(list(triangle_observed), aggregator_args))
+    n_horizons <- nrow(aggr_obs)
+    for (d in 1:n_horizons) {
+      indices_nowcast <- is.na(aggr_rt_obs)
+      indices_obs <- !is.na(aggr_obs)
+      masked_nowcast <- .apply_mask(aggr_nowcast, indices_nowcast, indices_obs)
+      masked_obs <- .apply_mask(aggr_obs, indices_nowcast, indices_obs)
+      exp_to_add[i,d] <- sum(masked_nowcast, na.rm = TRUE)
```
kaitejohnson, line 169: We always sum after taking the mask. I think I have messed this up and it should be only for the dth row (so its really a sum across columns only so will fix).

R/sample_predictions.R
```diff
@@ -46,6 +49,7 @@
 sample_prediction <- function(point_nowcast_matrix,
                               reporting_triangle,
                               dispersion,
+                              observation_model = "negative binomial",
```
seabbs, line 52: the pass in here should be a function i.e your fitting function and similar.

R/estimate_uncertainty.R
```diff
@@ -49,32 +57,39 @@
...
 #' disp_params_agg
 estimate_uncertainty <- function(
     pt_nowcast_matrices,
     trunc_reporting_triangles,
     retro_reporting_triangles,
     n = length(pt_nowcast_matrices),
-    fun_to_aggregate = sum,
-    k = 1) {
-  .validate_aggregation_function(fun_to_aggregate)
+    error_model = .fit_distrib,
```
seabbs, line 86: the input can't be an intenral funciton this is user facing

R/estimate_uncertainty.R
```diff
@@ -49,32 +57,39 @@
...
+    error_args = list(observation_model = "negative binomial"),
```
seabbs, line 87: I think the function above should have this as a arg via a partial function approach so we should't need this argument.

R/sample_predictions.R
kaitejohnson: Yes need to update this

R/estimate_uncertainty.R
```diff
@@ -49,32 +57,39 @@
...
+    aggregator = zoo::rollsum,
```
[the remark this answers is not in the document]
kaitejohnson: Ok -- wouldn't that require a `delay_aggregator` and `reference_time_aggregator` where `delay_aggregator = sum` in the current default case, `reference_time_aggregator = sum` in the default or `reference_time_aggregator = rollsum` in the k !=1 case?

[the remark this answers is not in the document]
seabbs: no I don't think so. You would pass in a function that does boht

R/estimate_uncertainty.R
kaitejohnson: > no function(...) fit_dist(..., option = blah blah)

I am not following this

R/estimate_uncertainty.R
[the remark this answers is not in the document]
kaitejohnson: What example are you referring to?

R/estimate_uncertainty.R
```diff
@@ -137,72 +147,83 @@ estimate_uncertainty <- function(
...
-        ), na.rm = TRUE),
-        na.rm = TRUE
+      masked_obs <- .apply_mask(
+        aggr_obs[row_number, ],
+        indices_nowcast,
+        indices_obs
       )
-    } # end loop over forecast horizons
-  } # end loop over retrospective nowcast times
+      exp_to_add[i, d] <- sum(masked_nowcast, na.rm = TRUE)
```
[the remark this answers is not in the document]
kaitejohnson: I think you are right that this does not need to be a for loop, you could just use the `.apply_mask` and then the rowSums. Will edit.

R/sample_distribution.R
```diff
@@ -0,0 +1,65 @@
...
+#' pred <- matrix(c(
+#'   3.7, 6.1,
+#'   5.2, 10.4
+#' ), byrow = 2, nrow = 2)
+#' sampled_preds <- sample_distribution(pred,
+#'   uncertainty_params = c(50, 100),
+#'   observation_model_name = "negative binomial"
+#' )
+#' sampled_preds
+sample_distribution <- function(pred,
```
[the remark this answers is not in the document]
kaitejohnson: Fair this probably isn't necessary since it doesn't rely on any of the aggregator stuff. I will remove.

R/estimate_uncertainty.R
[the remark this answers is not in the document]
seabbs: https://github.com/epinowcast/baselinenowcast/pull/208#discussion_r2236650187

NAMESPACE
```diff
@@ -21,7 +23,11 @@ importFrom(cli,cli_abort)
 importFrom(cli,cli_alert_danger)
 importFrom(cli,cli_alert_info)
 importFrom(cli,cli_warn)
+importFrom(stats,dgamma)
```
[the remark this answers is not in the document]
seabbs: you are adding more and more technical debit of your own reproducing them that would be the reason to think about it. 

At the min just have support for the nb in the package and then either just doc you can you fitdistr or have a suggests dep that loads in some kind of function like that that wraps it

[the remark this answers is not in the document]
seabbs: i.e like this: https://github.com/epinowcast/primarycensored/blob/89aca63fde7984704a86a44eabaf9b6be94eae78/R/fitdistdoublecens.R#L90

note the deps of the main package.

R/estimate_uncertainty.R
```diff
@@ -137,72 +147,83 @@ estimate_uncertainty <- function(
...
-      # Function to aggregate is always applied after the matrix has been
-      # summed across delays
-      exp_to_add[i, d] <- fun_to_aggregate(
-        rowSums(as.matrix(nowcast *
-          indices_nowcast * indices_observed), na.rm = TRUE),
-        na.rm = TRUE
+
+    # Apply the aggregation to the truncated observations, the nowcast,
+    # and the reporting triangle.
+    aggr_obs <- do.call(aggregator, c(list(trunc_matr_observed), aggregator_args))
```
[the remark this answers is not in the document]
seabbs: We discussed passing them as one function which would preclude doing it this way as you need the mask (unless you pass it in which you could do but it adds complexity). 

> the mask has to come after the

If this is unavoidable then the design pattern we have been discussing won't work so either have to ag args ro and col or integrate the mask as an input into the one agg function.

R/sample_distribution.R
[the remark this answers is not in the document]
seabbs: I'm not clear why that is what you have rather than fitting the natural parameters?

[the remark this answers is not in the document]
seabbs: The gam? Sorry my question was a bit stupid you need the mean as you have to model the mean via the point nowcast. All seems reasonable

R/estimate_uncertainty.R
```diff
@@ -166,23 +166,20 @@ estimate_uncertainty <- function(
...
+      aggr_nowcast[(max_t - n_possible_horizons + 1):max_t, ],
+      indices_nowcast,
+      indices_obs
+    )
+    masked_obs <- .apply_mask(
+      aggr_obs[(max_t - n_possible_horizons + 1):max_t, ],
+      indices_nowcast,
+      indices_obs
+    )
+    exp_to_add[i, ] <- rev(rowSums(masked_nowcast, na.rm = TRUE))
```
seabbs, line 181: This is the bit that should be a passed in function

seabbs:
Anything useful from me?

R/estimate_uncertainty.R
kaitejohnson: I think I asked this previously, but how do we do this without requiring two separate functions (aggregation across reference times and aggregation across delays)? This one is the aggregation across delays which we currently have removed as a choice for the user.

kaitejohnson: Right and I think we can't do the masking without the aggregation across reference times per my understanding...

vignettes/baselinenowcast.Rmd
```diff
@@ -419,7 +419,7 @@ Now that we have estimated the dispersion, we can generate a probabilistic nowca
 ```{r}
 nowcast_draws_df <- sample_nowcasts(
   point_nowcast_matrix, reporting_triangle,
-  dispersion = disp_params,
+  uncertainty_params = disp_params,
```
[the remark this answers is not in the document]
kaitejohnson: these are fine as disp bc they are negative binomial

NAMESPACE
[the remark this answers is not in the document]
kaitejohnson: Ok if I make a separate issue to use `fitdistr` with a suggests dependency?

kaitejohnson:
> Anything useful from me?

Just confirming the implementation of the two separate aggregator functions and args is along the lines of what you were thinking.

kaitejohnson:
> Can you summarise what from the last review you have decided to do and not do and why. From a skim it looks like a few comments are still open.

Yes I tried to summarise in the opening comment but:
- `estimate_uncertainty` and `sample_prediction` take in two separate aggregator functions (delay and ref time aggregators) and an error model
- separate function for applying the mask which happens after reference time aggregation before delay aggregation (my understanding is that this is the required order, as we need to know other ref times outside of the mask to do the aggregation)
- wrote a helper function for the error model which fits and samples a distribution for either a negative binomial, normal, or gamma observation model. This is necessary to support the gamma observation model, which would typically return two sets of parameters (shape and scale) and we need it to return one (the sd) since the mean is being set by the point nowcast

R/estimate_uncertainty.R
```diff
@@ -15,19 +15,28 @@
...
-#' @param k Integer indicating the number of reference times to apply the
-#'    `fun_to_aggregate` over to create target used to compute the nowcast
-#'    errors.
+#' @param error_model Function that ingests a matrix of observations and a
+#'     matrix of predictions and returns a vector that can be used to
+#'     apply uncertainty using the same error model. Default is
+#'     fit_distribution.
+#' @param error_args List of arguments needed for the specified error model.
+#'     Default is `list(observation_model_name = "negative_binomial").`
+#' @param ref_time_aggregator Function that operates along the rows (reference
```
seabbs, line 24: we should need no error args here pass the full function. The default should be a passthrough function. Then we can drop zoo as a dep

R/estimate_uncertainty.R
```diff
@@ -135,73 +153,147 @@ estimate_uncertainty <- function(
...
-      to_add_already_observed[i, d] <- fun_to_aggregate(
-        rowSums(as.matrix(
-          obs * indices_nowcast * indices_observed
-        ), na.rm = TRUE),
-        na.rm = TRUE
+    )
+    max_t <- nrow(aggr_obs)
+    # For each horizon, take the partial sum of the nowcasted and already
+    # observed components.
+    indices_nowcast <- is.na(
```
seabbs, line 226: both of these indices finders should be internal functions

R/fit_distribution.R
```diff
@@ -0,0 +1,121 @@
...
+#' ), byrow = 2, nrow = 2)
+#' pred <- matrix(c(
+#'   3.7, 6.1,
+#'   5.2, 10.4
+#' ), byrow = 2, nrow = 2)
+#' disp <- fit_distribution(obs, pred,
+#'   observation_model_name = "negative binomial"
+#' )
+#' disp
+fit_distribution <- function(obs,
```
seabbs, line 28: We don't want to add this if the plan is to remove it in favour of a suggests usage of fitdistrplus. In general functions with nested lists should be avoided (see previous comment).

I suggest you simplify here and write a simple fit_nb or similar function similar to what was implemented/in the design doc. 

the looping over dispersion params should really be a utility function takes in a fitting function and does the looping. Then we need a similar one for samples to apply this or we enforece it should always be a vector of length delay and add another utility function that maps the single value to a vector. 

In inpuit this would look like `error_model = function(...) fit_by_delay(error_model = fit_nb, ...)`

Or similar depending on how careful you want to be.

The same general advice holds here for the sampling. 

Use functions to do any mean mapping you need to do (can side bar this for later as adding more dists is imo not part of this PR) have these as pass in functions. Apply them where needed. 

So in the simple case we have here we would have a function called i.e sample_nb that expects a vector if we go with that as the standard or a either and a mapping function if we don't

seabbs:
See my specific comments but at a high level I think this has drifted from the original issue scope and design doc. I would suggest minimising the number of new features added here (especially as the design is unclear) and instead focus on the core refactor. For that we are still not hitting the original spec'd input idea and its not clear to me why. Essentially we just need to wrap any custom args in a function call and pass that in rather than using the do.call approach as this leads to a very error prone and clunky interface. See scoringutils and specificially the metrics for how we can do that.

kaitejohnson:
> See my specific comments but at a high level I think this has drifted from the original issue scope and design doc. I would suggest minimising the number of new features added here (especially as the design is unclear) and instead focus on the core refactor. For that we are still not hitting the original spec'd input idea and its not clear to me why. Essentially we just need to wrap any custom args in a function call and pass that in rather than using the do.call approach as this leads to a very error prone and clunky interface. See scoringutils and specificially the metrics for how we can do that.

 This? https://github.com/epiforecasts/scoringutils/blob/3a6216cdb593e28dbc06b02f94f1def9a817d211/R/score.R#L128
 
 and this? 
 https://github.com/epiforecasts/scoringutils/blob/main/R/metrics.R

PR #225, opened 2025-08-04
Issue 218: fix truncate triangles default

### LINKED ISSUE #218: Default on `truncate_triangles` needs to change for ragged triangles
**Describe the bug**
The current default for the `truncate_triangles` function assumes we want `nrow(reporting_triangle) - ncol(reporting_triangle) - 1` truncated triangles. However, this can be negative if the triangle is jagged, as can be seen in the reprex below. 

**To Reproduce**
```r
sim_delay_pmf <- c(0.1, 0.2, 0.3, 0.1, 0.1, 0.1)

# Generate counts for each reference date
counts <- c(150,
  160, 170, 200, 100
)

# Create a complete triangle based on the known delay PMF
complete_triangle <- lapply(counts, function(x) round(x * sim_delay_pmf))
complete_triangle <- do.call(rbind, complete_triangle)

ragged_triangle <- construct_triangle(
  complete_triangle,
 structure = 2
)

trunc_rts <- truncate_triangles(ragged_triangle, n = 2)
#> Error in truncate_triangles(ragged_triangle) : 
#> Assertion on 'n' failed: Element 1 is not >= 0.
```
This is because:
```r
nrow(reporting_triangle) - ncol(reporting_triangle) - 1
#> -2
```

**Expected behavior**
Again, this is an issue because ncol(reporting_triangle) - 1 is assumed to be the number of horizons I think....

I think it would be fine to just have the default be `nrow(reporting_triangle) - 1` even though we know that not all of these will be nowcastable? In general, this should be specified. 

**Additional context**

Again bug found while implementing weekday filter and debugging #217

kaitejohnson:
## Description

This PR closes #218. The default value for `truncate_triangles` of the number of triangles to produce was based on logic assuming a symmetric reporting triangle, and could actually be negative for ragged reporting triangles. 

I have changed the default for this function to be safer as it is set to be the maximum number of triangles for which a point nowcast can be generated, with the minimum number of rows being the number of horizons (rows with NAs) plus 1 (since we need one full row). 

- Adds a test on a ragged reporting triangle to ensure the default is appropriate. 

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Updated the default truncation behavior to better account for rows with missing values in reporting triangles.

* **Documentation**
  * Clarified the explanation of the default truncation parameter to improve user understanding.

* **Tests**
  * Expanded test coverage to validate input handling, default behavior, and ragged triangle structures.
  * Added tests for uncertainty estimation with varied input sizes and expected warnings.

### DIFF
--- a/tests/testthat/test-truncate_triangles.R
+++ b/tests/testthat/test-truncate_triangles.R
@@ -30,7 +30,8 @@ test_that("truncate_triangles returns correct number of truncated matrices with 
 })
 
 test_that("truncate_triangles default n calculation works correctly", {
-  expected_default <- nrow(test_triangle) - sum(is.na(rowSums(test_triangle))) - 1
+  expected_default <- nrow(test_triangle) -
+    sum(is.na(rowSums(test_triangle))) - 1
   result <- truncate_triangles(test_triangle)
   expect_length(result, expected_default)
 })


R/truncate_triangles.R
```diff
@@ -8,9 +8,9 @@
 #'
 #' @param n Integer indicating the number of retrospective
 #'   truncated triangles to be generated, always starting from the most
-#'   recent reference time. Default is to generate truncated matrices
-#'   for all triangles that would have a sufficient number of rows to generate
-#'   a nowcast from.
+#'   recent reference time. Default is to generate truncated matrices for
+#'   the maximum delay (number of columns - 1) or one less than the number of
```
seabbs, line 12: no indication of which it will pick. I would have thought that most users would expect it to do one for each row so perhaps we need some rational for this default as well

tests/testthat/test-truncate_triangles.R
```diff
@@ -30,7 +30,7 @@ test_that("truncate_triangles returns correct number of truncated matrices with
 })
 
 test_that("truncate_triangles default n calculation works correctly", {
```
seabbs, line 32: need an explicit test for the edge case where it uses the min

seabbs (commented):
Looks good some specific comments.

tests/testthat/test-truncate_triangles.R
kaitejohnson: will add

R/truncate_triangles.R
seabbs: Yes aI think this is preferable but not really my point here you don't say it uses the min

kaitejohnson: Will edit

tests/testthat/test-truncate_triangles.R
kaitejohnson: Have removed the minimum so it is just one set value now

PR #227, opened 2025-08-12
Issue 115: Clean up refactor

### LINKED ISSUE #115: Refactor estimate_dispersion

kaitejohnson:
## Description

This PR closes #115. Opening in a new branch because I will remove a lot of what is in #115 but sequentially... 

## Checklist


seabbs:
Hmm sorry is this a new draft pr for #115 ? Probably best to close the old PR then?

R/estimate_uncertainty.R
```diff
@@ -72,27 +70,17 @@
...
-    error_model = fit_distribution,
-    error_args = list(observation_model_name = "negative binomial"),
-    ref_time_aggregator = zoo::rollsum,
-    ref_time_aggregator_args = list(
-      k = 1,
-      align = "right"
-    ),
-    delay_aggregator = rowSums,
-    delay_aggregator_args = list(na.rm = TRUE)) {
+    error_model = function(x, mu) fit_nb(x, mu),
```
kaitejohnson, line 81: @seabbs Ok I have added this back in (you cannot just use `fitdistrplus::fit_dist()` from my understanding because it doesn't accept a vector of expectiations and observations. If we want to use `mle_dist()` we can replace the current `fit_nb` with this. 

My follow-up question is do we want to provide additional helper functions for other distributions, which I believe will still need these custom functions to pass in vectors of observations and predictions (similar to what is in the current `fit_distribution` function but I will remove it?

seabbs: the plan as we discussed it is to keep a renamed exported and improved fit_nb only with the suggestion in the previous PR beign to handle the looping over the data in a wrapper function that people can use if they want to get different dispersions see the previous PR for that suggestion 

The suggestion for fitdistrpllus from the previous PR was to make this its own issue, to plan to write a suggests wrapper that handles the formatting like primarycensored does and in the same essential way if we decide to go that way and to not add the custom fit distributions because of all the technical debt it adds

kaitejohnson: Ok will make the `fit_by_delay()` wrapper and open a new issue for supporting fitdistrplus

inst/make_hexsticker.R
```diff
@@ -127,8 +127,8 @@ retro_pt_nowcast_mat_list <- fill_triangles(
 
 # Estimate a vector of negative binomial dispersion parameters from the observations and estimates at each horizon
 disp_params <- estimate_uncertainty(
-  pt_nowcast_matrices = retro_pt_nowcast_mat_list,
-  trunc_reporting_triangles = trunc_rep_tri_list,
+  point_nowcast_matrices = retro_pt_nowcast_mat_list,
+  truncated_reporting_triangles = trunc_rep_tri_list,
   retro_reporting_triangles = retro_rep_tri_list,
   n = n_retrospective_nowcasts
 )
```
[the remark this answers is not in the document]
kaitejohnson: Default is negative binomial so ignoring

man/estimate_uncertainty.Rd
```diff
@@ -27,31 +28,38 @@ matrices with as many rows as available given the truncation.}
 \item{n}{Integer indicating the number of reporting matrices to use to
 estimate the dispersion parameters.}
```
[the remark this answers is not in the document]
kaitejohnson: fixed

man/estimate_uncertainty.Rd
```diff
@@ -60,30 +68,29 @@
...
+# Estimate dispersion parameters using default (negative binomial error
+# model on the sums
 disp_params <- estimate_uncertainty(
   point_nowcast_matrices = retro_nowcasts,
   truncated_reporting_triangles = trunc_rts,
-  retro_reporting_triangles = retro_rts,
-  n = 2
+  retro_reporting_triangles = retro_rts
 )
 disp_params
```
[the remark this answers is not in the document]
kaitejohnson: fixed

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,190 @@ estimate_uncertainty <- function(
   }
 
   n_possible_horizons <- sum(is.na(rowSums(list_of_rts[[1]])))
+  n_iters <- .calc_n_retro_nowcast_times(
+    list_of_obs,
+    n_possible_horizons,
+    ref_time_aggregator
+  )
```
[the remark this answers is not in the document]
kaitejohnson: added a check and a test for the check

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,190 @@
...
+  uncertainty_params <- c()
+  if (!is.null(obs)) {
+    for (i in seq_len(ncol(obs))) {
+      uncertainty_params[i] <- fun(obs[, i], pred[, i])
+    }
   }
 
-  return(disp_params)
+  return(uncertainty_params)
+}
```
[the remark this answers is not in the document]
kaitejohnson: Added this, note need to add tests for this function!

R/sample_predictions.R
```diff
@@ -39,47 +43,62 @@
...
-                              dispersion,
-                              fun_to_aggregate = sum,
-                              k = 1) {
-  .validate_aggregation_function(fun_to_aggregate)
-  if (length(dispersion) > nrow(point_nowcast_matrix)) {
+sample_prediction <- function(
+    point_nowcast_matrix,
+    reporting_triangle,
+    uncertainty_params,
+    error_model = function(pred, params) sample_nb(pred, params),
```
seabbs, line 53: pass sample_nb directly wrapper doesn't do anything

R/sample_predictions.R
```diff
@@ -163,30 +181,40 @@ combine_obs_with_pred <- function(predicted_counts,
...
-                               draws = 1000,
-                               fun_to_aggregate = sum,
-                               k = 1) {
+sample_predictions <- function(
+    point_nowcast_matrix,
+    reporting_triangle,
+    uncertainty_params,
+    draws = 1000,
+    error_model = function(pred, params) sample_nb(pred, params),
+    ref_time_aggregator = function(x) identity(x),
```
seabbs, line 193: As these are now  matrix operations are we confident ref_time and delay is the right name? They are actually linked to when they occur within the wider operation right? I don't have concise thoughts here but worth a review I think.

R/sample_predictions.R
```diff
@@ -224,23 +251,29 @@ sample_predictions <- function(point_nowcast_matrix,
...
-sample_nowcast <- function(point_nowcast_matrix,
-                           reporting_triangle,
-                           dispersion,
-                           ...) {
+sample_nowcast <- function(
+    point_nowcast_matrix,
+    reporting_triangle,
+    uncertainty_params,
+    error_model = function(pred, params) sample_nb(pred, params),
+    ref_time_aggregator = function(x) identity(x),
```
seabbs, line 259: all the same points about pass through complexity to be minimised

R/sample_predictions.R
kaitejohnson: Specifically, what I tried was : `error_model = sample_nb()` where the function args all had defaults (sometimes NULL)

R/sample_predictions.R
```diff
@@ -3,16 +3,20 @@
 #' @param point_nowcast_matrix Matrix of point nowcast predictions and
 #'   observations, with rows representing the reference times and columns
 #'   representing the delays.
-#' @param dispersion Vector of dispersion parameters indexed by horizon from
-#'  minus one to the maximum delay.
+#' @param uncertainty_params Vector of uncertainty parameters ordered from
+#'   horizon 1 to the maximum horizon. Note that these will be reversed
+#'   internally to match the ordering of the `point_nowcast_matrix` (where
+#'   a horizon of 1 is the last entry).
+#' @param error_model Function that ingests a vector or matrix of predictions
```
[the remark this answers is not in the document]
seabbs: The required output should also be documented but harder to check this without overheards. in primarycensorewd I provide a utility people can use themsevles to check their inputs but htinking more about this is its own issue. Ithink

R/sample_predictions.R
```diff
@@ -224,23 +251,29 @@ sample_predictions <- function(point_nowcast_matrix,
...
 #' nowcast_draw
-sample_nowcast <- function(point_nowcast_matrix,
-                           reporting_triangle,
-                           dispersion,
-                           ...) {
+sample_nowcast <- function(
+    point_nowcast_matrix,
+    reporting_triangle,
+    uncertainty_params,
+    error_model = function(pred, params) sample_nb(pred, params),
```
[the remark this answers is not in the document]
kaitejohnson: Again just made them explicit bc I had to -- I must still be missing something bc I couldn't just write `sample_nb()`

R/sample_predictions.R
```diff
@@ -275,23 +307,36 @@ sample_nowcast <- function(point_nowcast_matrix,
...
-                            reporting_triangle,
-                            dispersion,
-                            draws = 1000,
-                            ...) {
+sample_nowcasts <- function(
+    point_nowcast_matrix,
+    reporting_triangle,
+    uncertainty_params,
+    draws = 1000,
+    error_model = function(pred, params) sample_nb(pred, params),
```
seabbs, line 315: here is a classic case where ... was used for a reason i.e because we only want optional args in the singular function in general. Reflexct. We don't really want functions that have more than 6 args as a rule of thumb unless we really really can't avoid it. It suggests we are trying to do way to much. 

On that note we might want to consider using a constructor function instead i.e

uncertainty_opts(error, ref_time, delay, samople) and then pass that around it we had one for negbin i.e specialsied it could have nb_unertainty_obs(ref, delay) and we assign fit and sample internally. 

I think this is likely a good idea but for another issue  so here just reflect on if we should really be adding args

R/sample_predictions.R
```diff
@@ -300,3 +345,31 @@ sample_nowcasts <- function(point_nowcast_matrix,
   draws_df <- Reduce(rbind, draws_df_list)
   return(draws_df)
 }
+#' Sample from negative binomial model given a set of predictions
+#'
+#' @param pred Vector of predictions. Default is `NULL`
+#' @param uncertainty_params Vector of uncertainty parameters. Default is
+#'     `NULL`.
```
seabbs, line 352: why null is this pattern documented somewhere i.e where it is needed as an input

seabbs (commented):
Overall. Comments are about docs, names and points of reflection

R/sample_predictions.R
```diff
@@ -300,3 +345,31 @@ sample_nowcasts <- function(point_nowcast_matrix,
...
+#' @returns `sampled_pred` Object of the same dimensions as `pred` representing
+#'    a single draw from the negative binomial distribution
+#'    with the specified `uncertainty params`.
+#' @examples
+#' pred <- c(3.2, 4.6)
+#' sampled_preds <- sample_nb(pred,
+#'   uncertainty_params = c(50, 100)
+#' )
+#' sampled_preds
+sample_nb <- function(pred = NULL, uncertainty_params = NULL) {
```
seabbs, line 364: these should likely go in their own file especially if there are going to be more of them at some point

R/sample_predictions.R
```diff
@@ -107,22 +128,19 @@ sample_prediction <- function(point_nowcast_matrix,
...
-    fun_to_aggregate,
-    fill = NA,
-    align = "right"
-  )
-  return(obs_counts + predicted_counts)
+#' combine_obs_with_pred(pred_counts,
+#'   reporting_triangle,
+#'   ref_time_aggregator = function(x) zoo::rollsum(x, k = 2, align = "right")
+#' )
+combine_obs_with_pred <- function(
```
seabbs, line 136: just in general there is way too much functionality living in utils suggests issue for resort

R/estimate_uncertainty.R
```diff
@@ -41,40 +47,39 @@
...
 #' disp_params_agg
 estimate_uncertainty <- function(
     point_nowcast_matrices,
     truncated_reporting_triangles,
     retro_reporting_triangles,
     n = length(point_nowcast_matrices),
-    fun_to_aggregate = sum,
-    k = 1) {
-  .validate_aggregation_function(fun_to_aggregate)
+    error_model = function(obs, pred) fit_by_horizon(obs = obs, pred = pred),
```
seabbs, line 80: fit_by_horizon should take an error model as an arg as it is in reality as fancy for loop. then we can begin to compose function inputs which is the point of this design.

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,211 @@ estimate_uncertainty <- function(
...
-        na.rm = TRUE
-      )
-      to_add_already_observed[i, d] <- fun_to_aggregate(
-        rowSums(as.matrix(
-          obs * indices_nowcast * indices_observed
-        ), na.rm = TRUE),
-        na.rm = TRUE
+
+    # Apply the aggregation to the truncated observations, the nowcast,
+    # and the reporting triangle.
```
seabbs, line 186: see later comment on if this is the right name

R/estimate_uncertainty.R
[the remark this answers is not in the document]
kaitejohnson: The validation is already in the code, I have added another test for a bad aggregator

tests/testthat/test-sample_predictions.R
```diff
@@ -148,3 +151,24 @@ test_that("sample_predictions works with different number of draws", {
...
+    regexp = "Vector of uncertainty parameter is less than the number "
+  )
+
+  expect_error(sample_predictions(
+    point_nowcast_matrix,
+    reporting_triangle,
+    c(dispersion, rep(3, 3)),
+    draws = 10
+  ))
+})
```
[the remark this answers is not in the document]
kaitejohnson: added

R/estimate_uncertainty.R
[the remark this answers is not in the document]
kaitejohnson: added

R/estimate_uncertainty.R
kaitejohnson: I think my preference is to keep ref time aggregator and delay aggregator. I changed the argument to `estimate_uncertainty` to `uncertainty_model` as it is about how you are estimating the uncertainty from the matrices rather than what the specific type of observation model it is that you are using. 

`fit_by_horizon()` then takes in the `observation_model` (which is `fit_nb`).

R/estimate_uncertainty.R
kaitejohnson: Yup have changed this (but am using `observation_model`)

R/sample_predictions.R
kaitejohnson: I changed all the plural functions to use ...

R/sample_predictions.R
```diff
@@ -88,7 +107,9 @@ sample_prediction <- function(point_nowcast_matrix,
 #' time and adds them to the predicted counts to form a single draw of the
 #' nowcast for the final counts by reference time.
 #'
-#' @param predicted_counts Vector of predicted counts at each reference time
+#' @param predicted_counts Vector of predicted counts at each reference time.
+#'    Note that if using an aggregator function, this is assumed to have
```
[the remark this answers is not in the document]
kaitejohnson: Fixed docs. Can add a warning if not using defaults for both delay and reference time aggregators?

R/sample_predictions.R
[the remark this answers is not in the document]
kaitejohnson: Have updated docs, will make separate issue to verify outputs

R/sample_predictions.R
```diff
@@ -163,30 +181,40 @@ combine_obs_with_pred <- function(predicted_counts,
...
-
   reference_times <- seq_len(nrow(point_nowcast_matrix))
 
   draws_df_list <- lapply(seq_len(draws), function(i) {
     pred_counts <- sample_prediction(
       point_nowcast_matrix,
       reporting_triangle,
-      dispersion
+      uncertainty_params,
+      error_model,
```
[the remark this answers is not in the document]
kaitejohnson: Have changed this to `observation_model`. Now in `fit_by_horizon` we expect an observation model which is by default `fit_nb()` and in `sample_predictions` we expect an observation model which is by default `sample_nb`. Both fit and sample `nb` ingest vectors of either obs and preds and return a single uncertainty parameter or preds and uncertainty and return a vector of samples. 

I think this is the most straightforward way to do this -- though am noting that in `sample_prediction` we still have a requirement that the uncertainty parameters are vectors that pertain to the horizons of the nowcast -- I think we should make a separate issue to make this more general (it would have to remove the construction of the matrices where horizons are columns and rows are retrospective nowcast times and add something more complicated to track an arbitrary set of dimensions...

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,211 @@ estimate_uncertainty <- function(
...
+#'     4.2, 5.2, 1.8,
+#'     0.7, 3.5, 3.4,
+#'     7.3, 4.1, 1.2
+#'   ),
+#'   nrow = 3,
+#'   byrow = TRUE
+#' )
+#' disp <- fit_by_horizon(obs = obs, pred = pred)
+#' disp
+fit_by_horizon <- function(fun = function(x, mu) fit_nb(x, mu),
```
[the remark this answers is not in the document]
kaitejohnson: have fixed I think.

R/estimate_uncertainty.R
seabbs: why?

R/sample_predictions.R
seabbs: error_model = sample_nb() this is calling not passing the function

R/estimate_uncertainty.R
kaitejohnson: Why what? Why keep ref and delay? Because thats easier to remember then knowing ref = row and delay = column.

Why `uncertainty_model`? Because it is a description of how we are computing uncertainty not the observation model we are assuming between the predictions and observations -- maybe should be `uncertainty_operation` or `uncertainty_function`?

man/sample_prediction.Rd
```diff
@@ -69,8 +78,7 @@ nowcast_pred_draw_agg <- sample_prediction(
   point_nowcast_matrix,
   reporting_triangle,
   disp,
-  fun_to_aggregate = sum,
-  k = 2
+  ref_time_aggregator = function(x) zoo::rollsum(x, k = 2, align = "right")
```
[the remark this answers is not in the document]
kaitejohnson: have added

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,211 @@ estimate_uncertainty <- function(
...
+    # Apply the aggregation to the truncated observations, the nowcast,
+    # and the reporting triangle.
+    aggr_obs <- ref_time_aggregator(trunc_matr_observed)
+    aggr_nowcast <- ref_time_aggregator(nowcast_i)
+    aggr_rt_obs <- ref_time_aggregator(triangle_observed)
+
+    # For each horizon, take the partial sum of the nowcasted and already
+    # observed components.
+
+    indices_nowcast <- is.na(aggr_rt_obs |>
```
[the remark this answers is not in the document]
kaitejohnson: have fixed

R/sample_predictions.R
kaitejohnson: Have made a new issue https://github.com/epinowcast/baselinenowcast/issues/230

R/estimate_uncertainty.R
```diff
@@ -135,73 +140,211 @@ estimate_uncertainty <- function(
...
+fit_by_horizon <- function(fun = function(x, mu) fit_nb(x, mu),
+                           obs = NULL,
+                           pred = NULL) {
+  if (is.null(obs) || is.null(pred)) {
+    return(numeric(0))
   }
 
-  return(disp_params)
+  # Ensure obs and pred have the same dimensions
+  if (!identical(dim(obs), dim(pred))) {
```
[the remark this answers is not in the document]
kaitejohnson: I added a checker function outside the `fit_by_horizon` within `estimate_uncertainty`

R/estimate_uncertainty.R
```diff
@@ -135,73 +145,225 @@ estimate_uncertainty <- function(
...
   n_possible_horizons <- sum(is.na(rowSums(list_of_rts[[1]])))
+  n_iters <- .calc_n_retro_nowcast_times(
+    list_of_obs,
+    n_possible_horizons,
+    ref_time_aggregator
+  )
+  if (n_iters == 0) {
+    cli_abort("No valid retrospective nowcast times after reference time aggregation.") # nolint
+  }
+
```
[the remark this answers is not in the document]
kaitejohnson: It should always be the latest

seabbs (approved):

seabbs:
Sorry I don't understand. 

My point was what is the reason fit_by_horizon doesn't have uncertainty_model as its input arg i.e. to make it composable and what is the reason to use a name i.e. observation_model to mean something else at the sampling stage (here it means sample not fit).

kaitejohnson:
I went with because I think these all should be different and indicate whether they are fitting or sampling. 

```r
estimate_uncertainty(..., uncertainty_model = fit_by_horizon)
fit_by_horizon(..., fit_model = fit_nb)
sample_prediction(..., uncertainty_sampler = sample_nb)
```

I think we will eventually want this:
```r
estimate_uncertainty(..., uncertainty_model = fit_by_horizon)
fit_by_horizon(..., fit_model = fit_nb)
sample_prediction(..., uncertainty_sampler = sample_by_horizon)
sample_by_horizon(..., sample_model = sample_nb)
```

But for now this is a bit of a large refactor to make it flexible to handle something that doesn't apply uncertainty by horizon necessarily.

Issue #193, opened 2025-07-15
Implementation of user interface


kaitejohnson:
This issue is an updated design document detailing the specifics the baselinenowcast user interface . The suggested original design is outlined in #86. 

## Current State

The `baselinenowcast` package provides low-level modular functions for nowcasting epidemiological data with reporting delays. The current workflow requires users to:

1. Manually convert their data to reporting triangle matrices
2. Call multiple functions sequentially to estimate delays and generate nowcasts
3. Handle all input/output conversions between data.frames and matrices
4. Execute multi-step workflows to properly account for uncertainty

### High-Level Interface Function: `baselinenowcast()`

Create a single user-facing function that serves as the main entry point:
```r
baselinenowcast(
  data,                      # Will detect input type using S3 methods (either `reporting_triangle` or `reporting_triangles`)
  strata = NULL,             # Column names for stratification e.g. location, age group, weekday
  strata_sharing = NULL,     # Character: "delay", "uncertainty", both, or empty
  include_draws = TRUE,      # Whether to include uncertainty
  scale_factor = 3,   # Same default as `allocate_reference_times` for the number of reference times to use for all training
  prop_delay = 0.5,    # Same default as `allocate_reference_times` for the number of reference times to use for delay
  draws = 100,               # Number of draws if include_draws=TRUE
  verbose = FALSE,           # Whether to provide step-by-step output *Note I don't know what we mean by this*
  ...                        # Passed to estimate_uncertainty()
)
```
Key features of the wrapper function:
- implements default behaviour on a reporting triangle object and strata (individually as a default) 
- enables nowcasting multiple strata (e.g. locations, age groups, weekdays)
- applies strata sharing if this is desired (note, to share across some strata but not others, users will need to interact with low level functions and pass things around accordingly, this is an "all or nothing" sharing across strata boolean 
- allows users to specify the scale factor on the max delay used for estimation and the proportion used for delay or uses defaults if nothing is passed 

Step by step it would:
1. Accept either `reporting_triangle` class object (which contains a reporting triangle dataframe with additional attributes such as reference dates and strata) or a `reporting_triangles` class object (which contains a list of `reporting_triangle`s)
2. Compute shared delays and uncertainty parameters if needed
3. Compute the training volume allocation for each strata 
4. Execute the nowcasting workflow separately on each reporting triangle, combine with reference dates and metadata on strata       
5. If `reporting_triangles` method, bind the dataframe across strata
[...]

PR #242, opened 2025-10-07
Issue 241: `baselinenowcast.reporting_triangle()` method

### LINKED ISSUE #241: `baselinenowcast.reporting_triangle()` method
As described in #193

kaitejohnson:
## Description

This PR closes #241. It aims to write the single `reporting_triangle` method for `baselinenowcast`, which produces a `nowcast_df` class object (which is also a data.frame). 

Note, in doing so, it changes the `reporting_triangle()` inputs as we want to distinguish between what we need to recreate a `nowcast_df`, which is both the title of the grouping variable and the value of the grouping variable (so we need a named list with single entries). This named list is called `strata_map`.

 This is separate from what we want form the user in the `as_reporting_triangle.data.frame()` method, where we want to know the column name in their data.frame which corresponds to the strata metadata. This is called `strata`.

Additional tasks: 
- added tests for no uncertainty to make sure that works
- add tests which ensure that passing a separate delay /uncertainty is different from using the one from the triangle itself

## Checklist

## Summary by CodeRabbit

- New Features
  - Introduced baselinenowcast for reporting triangles to produce nowcast data frames, with optional uncertainty draws.
  - Added new_nowcast_df to assemble nowcast outputs with reference dates and strata.
- Refactor
  - Replaced strata with strata_map across reporting triangle creation and structure.
- Documentation
  - Added docs for baselinenowcast, its reporting_triangle method, validation helpers, new_nowcast_df, and updated reporting triangle references to strata_map.
- Tests
  - Added comprehensive tests for baselinenowcast and new_nowcast_df; updated existing tests to use strata_map.
- Style/Validation
  - Enhanced input validation for delay distributions and uncertainty parameters.

### DIFF
--- a/R/baselinenowcast.R
+++ b/R/baselinenowcast.R
@@ -111,8 +111,10 @@ baselinenowcast.reporting_triangle <- function(
       time = seq_len(nrow(pt_nowcast)),
       pred_count = rowSums(pt_nowcast)
     )
+    nowcast_df$draw <- 1
     result_df <- new_baselinenowcast_df(nowcast_df,
-      reference_dates = data$reference_dates
+      reference_dates = data$reference_dates,
+      output_type = output_type
     )
     return(result_df)
   }
@@ -142,7 +144,8 @@ baselinenowcast.reporting_triangle <- function(
   }
 
   result_df <- new_baselinenowcast_df(nowcast_df,
-    reference_dates = data$reference_dates
+    reference_dates = data$reference_dates,
+    output_type = output_type
   )
 
   return(result_df)
--- a/R/baselinenowcast_df-class.R
+++ b/R/baselinenowcast_df-class.R
@@ -16,7 +16,11 @@
 #'  \item{pred_count}{Numeric indicating the estimated total counts aggregated
 #'    across delays at each reference date.}
 #'  \item{draw}{Integer indexing the sample from the probabilistic nowcast
-#'    distribution. Only present if `output_type = "samples"`.}
+#'    distribution. If `output_type = "point"`, this will be set to 1.}
+#'  \item{output_type}{Character string indicating whether the `pred_count`
+#'   represents a probabilistic draw from the observation model indicated by
+#'   `"samples"` or whether the `pred_count` is a point estimate indicated by
+#'   `"point"`.}
 #' }
 #' See the corresponding \code{\link{reporting_triangle}} and
 #' \code{\link{baselinenowcast}} function
@@ -33,11 +37,14 @@ NULL
 #'  (`pred_count`), and optionally the draw number (`draw`).
 #' @param reference_dates Vector of reference dates corresponding to the
 #'    reference times in the `baselinenowcast_df`.
+#' @inheritParams baselinenowcast
 #'
 #' @returns An object of class \code{\link{baselinenowcast_df}}
 #' @export
 new_baselinenowcast_df <- function(baselinenowcast_df,
-                                   reference_dates) {
+                                   reference_dates,
+                                   output_type) {
+  assert_choice(output_type, choices = c("samples", "point"))
   spine_df <- data.frame(
     reference_date = reference_dates,
     time = seq_along(reference_dates)
@@ -50,6 +57,7 @@ new_baselinenowcast_df <- function(baselinenowcast_df,
   )
 
   baselinenowcast_df_dates$time <- NULL
+  baselinenowcast_df_dates$output_type <- output_type
 
   result <- structure(
     data.frame(baselinenowcast_df_dates),
--- a/man/baselinenowcast_df-class.Rd
+++ b/man/baselinenowcast_df-class.Rd
@@ -18,7 +18,11 @@ nowcast.}
 \item{pred_count}{Numeric indicating the estimated total counts aggregated
 across delays at each reference date.}
 \item{draw}{Integer indexing the sample from the probabilistic nowcast
-distribution. Only present if \code{output_type = "samples"}.}
+distribution. If \code{output_type = "point"}, this will be set to 1.}
+\item{output_type}{Character string indicating whether the \code{pred_count}
+represents a probabilistic draw from the observation model indicated by
+\code{"samples"} or whether the \code{pred_count} is a point estimate indicated by
+\code{"point"}.}
 }
 See the corresponding \code{\link{reporting_triangle}} and
 \code{\link{baselinenowcast}} function
--- a/man/new_baselinenowcast_df.Rd
+++ b/man/new_baselinenowcast_df.Rd
@@ -4,7 +4,7 @@
 \alias{new_baselinenowcast_df}
 \title{Combine data from a nowcast dataframe, strata, and reference dates}
 \usage{
-new_baselinenowcast_df(baselinenowcast_df, reference_dates)
+new_baselinenowcast_df(baselinenowcast_df, reference_dates, output_type)
 }
 \arguments{
 \item{baselinenowcast_df}{Data.frame containing information for multiple
@@ -13,6 +13,10 @@ draws with columns for the reference time (\code{time}), the predicted counts
 
 \item{reference_dates}{Vector of reference dates corresponding to the
 reference times in the \code{baselinenowcast_df}.}
+
+\item{output_type}{Character string indicating whether the output should be
+samples (\code{"samples"})from the estimate with full uncertainty or whether to
+return the point estimate (\code{"point"}). Default is \code{"samples"}.}
 }
 \value{
 An object of class \code{\link{baselinenowcast_df}}
--- a/tests/testthat/test-baselinenowcast.R
+++ b/tests/testthat/test-baselinenowcast.R
@@ -5,19 +5,21 @@ rep_tri <- as_reporting_triangle(
   max_delay = 25,
   strata = "00+"
 )
-expected_cols <- c("pred_count", "draw", "reference_date")
+expected_cols <- c("pred_count", "draw", "reference_date", "output_type")
 test_that("baselinenowcast.reporting_triangle() works as expected", {
   nowcast_df <- baselinenowcast(rep_tri, draws = 100)
   expect_s3_class(nowcast_df, "data.frame")
   expect_s3_class(nowcast_df, "baselinenowcast_df")
   expect_true(all(expected_cols %in% colnames(nowcast_df)))
+  expect_identical(nowcast_df$output_type[1], "samples")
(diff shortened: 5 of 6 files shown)


kaitejohnson:
To make this clear in the design I think we have two options, the first is what I have and is consistent with the current `as_reporting_triangle()` methods in main:

## Option 1: Keep `strata` and extract it from the `data.frame` in the `as_reporting_triangle.data.frame()` method
- `reporting_triangle` object: list with reporting triangle matrix, reference dates, and a list for strata (grouping variable and the actual entry). We are strict and ensure that strata pertains to a single variable in the `data.frame` method . 
- `reporting_triangles` object: list of `reporting_triangle`s, no additional metadata fields
- `baselinenowcast.reporting_triangle()` -> `nowcast_df` this is a dataframe with the nowcast + columns for the strata (we ensure single strata)
- `baselinenwocast.reporting_triangles()` -> `nowcasts_df` this is a dataframe that binds together `nowcast_df`s

**Advantages of this approach:** 
- We can check that when the user calls `as_reporting_triangle.data.frame()` that only a single strata pertains to that reporting triangle (ensuring the length of the unique strata columns specified is 1)
- The `baselinenowcast.reporting_triangles()` method is a very thin wrapper looping through the `reporting_triangle`s and calling `baselinenowcast.reporting_triangle()`
- The information in the `data.frame` in the `as_reporting_triangle.data.frame()` method is maintained for the user if they specify the `strata` (and it is not if they don't) 

**Disadvantages:**
- `strata` is carried into the `reporting_triangle` object, when in theory for a single nowcast, the user could just pass in a data.frame, lose the strata information, and join it back themselves at each iteration. However, the user could always set `strata` as NULL if they wanted to do this and that would just be the default. 

## Option 2: Remove strata from `reporting_triangle` and keep as metadata in `reporting_triangles()`
- `reporting_triangle` object: list with reporting triangle matrix and reference dates
- `reporting_triangles` object: list of `reporting_triangle`s, separately a `strata` list which must be a named list where the entries are vectors of the length of the number of `reporting_triangle`s, and whose order must match the order of the `reporting_triangle`s
- `baselinenowcast.reporting_triangle()` --> `nowcast_df` this is a dataframe with the nowcast and no additional columns
- `baselinenowcast.reporting_triangles()` --> `nowcasts_df` this is a dataframe that binds together the `nowcast_df` and assigned the `strata` from `reporting_triangles` accordingly (assuming order is the same) 

**Advantages**
- doesn't unnecessarily pass around a `strata` argument in `reporting_triangle` or `nowcast_df`

**Disadvantages**
- could be error prone because of the separation between the information pertaining to a single `reporting_triangle` 

Ultimately, I think either are reasonable, and if you have strong preferences for Option 2 I will defer to you given I must be having a blindspot in terms of the potential technical debt Option 1 incurs. If my Option 2 is not what you were envisioning though, it would be helpful to have mapped out what you have in mind.

kaitejohnson:
> It accepts a default of null?

Yes. Did you not just say you don't want the low level `reporting_triangle` and its associated methods to have any tracking of strata?

kaitejohnson:
> My suggestion in #193 is to use a light dataframe wrapper to store multiple triangles so you have `reporting_triangles_df` for singular in a df or multiple in a df with metadata and then `reporting_triangle` for the single case which is a light matrix wrapper.

Was referring to this comment? If you were going to write a `baselinenowcast.reporting_triangles_df` wouldn't you need to create a `reporting_triangles_df` class object (e.g. the light dataframe wrapper with attributes)?

R/baselinenowcast.R
```diff
@@ -0,0 +1,212 @@
...
+    nowcast_df <- sample_nowcasts(
+      point_nowcast_matrix = pt_nowcast,
+      reporting_triangle = tri,
+      uncertainty_params = uncertainty_params,
+      draws = draws,
+      uncertainty_sampler = uncertainty_sampler,
+      ...
+    )
+  }
+
```
seabbs:
```suggestion
```

R/baselinenowcast.R
```diff
@@ -0,0 +1,212 @@
...
+  if (any(dups)) {
+    cli_abort(
+      message = c(
+        "Data contains multiple `reference_date`s", # nolint
+        "x" = "Found {sum(dups)} duplicate `reference_date`{?s}", # nolint
+        "i" = "`nowcast_df` objects should only contain a single estimate for each reference date." # nolint
+      )
+    )
+  }
+
```
seabbs: space

R/nowcast_df-class.R
```diff
@@ -0,0 +1,30 @@
...
+#' \describe{
+#'  \item{reporting_triangle_matrix}{Matrix with which rows are reference
+#'    times and columns are delays and
+#'    entries are incident cases at each reference time and delay.}
+#'  \item{reference_date}{Dates corresponding to the reference times of the
+#'  nowcast.}
+#'  \item{time}{Integer indicating the time index}
+#'  \item{pred_count}{Numeric indicating the estimated total counts aggregated
+#'    across delays at each reference date.}
+#'  \item{draw}{Integer indexing the sample from the probabilistic nowcast
```
seabbs: you could save yourself some future pain in plotting etc but having this always be there but being 1 for point

R/nowcast_df-class.R
```diff
@@ -0,0 +1,30 @@
...
+#'
+#' @description
+#' A `nowcast_df` object which contains point or probabilistic nowcasts
+#'   alongside reference dates and any additional metadata, in tidy data format.
+#'   Nowcasts are presented aggregated across delays, by reference date.
+#'
+#' @section Structure:
+#' A `nowcast_df` is a data.frame with the following columns:
+#' \describe{
+#'  \item{reporting_triangle_matrix}{Matrix with which rows are reference
```
seabbs: Is this the case? I don't see it in the code above.

tests/testthat/test-new_nowcast_df.R
```diff
@@ -0,0 +1,81 @@
...
+
+  expect_s3_class(new_df, "data.frame")
+  expect_identical(nrow(new_df), 30L)
+  expect_true(all(expected_cols %in% colnames(new_df)))
+
+  # time corresponds to reference dates starting at 1
+  expected_dates <- min(ref_dates) + lubridate::days(unique(data_df$time) - 1)
+  expect_equal(expected_dates, sort(unique(new_df$reference_date))) # nolint
+
+  # no additional columns if no strata are specified
```
seabbs: out of date?

seabbs (approved):
Close. Minor issues.

R/baselinenowcast.R
```diff
@@ -0,0 +1,212 @@
...
+      time = seq_len(nrow(pt_nowcast)),
+      pred_count = rowSums(pt_nowcast)
+    )
+    if (!is.null(uncertainty_params)) {
+      cli_warn(
+        message =
+          "`uncertainty_params` passed in but point estimate was specified as an output type. `uncertainty params` will not be used." # nolint
+      )
+    }
+  } else if (is.null(uncertainty_params)) {
```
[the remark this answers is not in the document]
kaitejohnson: Yes so the slight issue, and up to you whether you care, is that we don't have an intermediate function that goes from a point nowcast, reporting triangle, and the allocations and returns only the uncertainty parameters. 

I could do as you say but it would at this point, require the more verbose code in which we truncate, construct, and nowcast the retrospective triangles and then pass to `estimate_uncertainty` (and then we could just call `sample_nowcasts()` once).

R/nowcast_df-class.R
kaitejohnson: A bit semantics but 1 draw from the observation model is different from the point estimate, so I feel like this might be misleading (e.g. in the frowned upon case where a user wants 1 probabilistic draw, the output would like identical to the point estimate in terms of metadata structure but it would not be).

We could add a column `output_type` and have that be `point` or `samples` if we want to do this, then for `draw` either have 1 or NULL for the point estimate.

R/baselinenowcast.R
```diff
@@ -0,0 +1,212 @@
...
+#'    triangle. See documentation for the arguments of this function which
+#'    can be used to set the model specifications (things like number of
+#'    reference times for delay and uncertainty estimation, the observation
+#'    model, etc.).
+#'
+#' @param data `reporting_triangle` class object to be nowcasted.
+#' @param delay_pmf Vector of delays assumed to be indexed starting at the
+#'   first delay column in `data$reporting_triangle_matrix`. Default is to
+#'   estimate from the reporting triangle matrix in `data`,
+#'   `estimate_delay(data$reporting_triangle_matrix)`.
```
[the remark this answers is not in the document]
kaitejohnson: And then set this internally

NEWS.md
```diff
@@ -1,5 +1,6 @@
 # baselinenowcast 0.0.0.1000
 
+-   Add a `baselinenowcast.reporting_triangle()` method which ingests a `reporting_triangle` object and produces a `nowcast_df` object which is a data.frame with nowcasts and associated metadata.
```
seabbs: out of date

R/baselinenowcast.R
```diff
@@ -0,0 +1,151 @@
...
+#'   data = data_as_of_df,
+#'   max_delay = 25
+#' )
+#' nowcast_df <- baselinenowcast(rep_tri)
+#' nowcast_df
+baselinenowcast.reporting_triangle <- function(
+    data,
+    scale_factor = 3,
+    prop_delay = 0.5,
+    output_type = "samples",
```
seabbs: doesn't match base method above

R/baselinenowcast.R
```diff
@@ -0,0 +1,151 @@
...
+  .validate_delay(tri, delay_pmf)
+
+  pt_nowcast <- apply_delay(tri, delay_pmf)
+
+  if (output_type == "point") {
+    nowcast_df <- data.frame(
+      time = seq_len(nrow(pt_nowcast)),
+      pred_count = rowSums(pt_nowcast)
+    )
+    if (!is.null(uncertainty_params)) {
```
seabbs: I really don't think you need this edge case check. I would remove and add if you get user reports of an issue.

seabbs (changes requested):
Missing description for the baselinenowcast generic entry point. I've nuked the comments as they are nearly verbatim the names of the functions and so don't imo adding anything except length. If you do find them more than marginally useful feel very free to leave in

R/baselinenowcast.R
kaitejohnson: made them both the same

PR #271, opened 2025-10-17
Issue 267: Improve warnings in `reporting_triangle` object and error if trying to `nowcast`

### LINKED ISSUE #267: Report dates after references dates results in an error in `structure` instead of a warning
**Describe the bug**
If the user passes in a dataframe with reference and report dates to `as_reporting_triangle.data.frame()` and there are report dates after the latest reference date through the maximum delay, it should just warn about this (or error if there is nothing to nowcast). 
 
Instead it errors because the first structure element is not greater than or equal to 1. 

**To Reproduce**
```r
library(tidyr)
data <- expand_grid(
  reference_date = seq(as.Date("2021-04-01"), as.Date("2021-04-30"), by = "day"),
  report_date = seq(as.Date("2021-04-01"), as.Date("2021-05-15"), by = "day")
) |>
  mutate(count = 5)

rep_tri <- as_reporting_triangle(data, max_delay = 10)

#> Error in .validate_rep_tri_args(reporting_triangle_matrix, reference_dates,  : 
#> Assertion on 'structure' failed: Element 1 is not >= 1.
#> In addition: Warning message:
#> The dataframe contains report dates beyond the final reference date. 
```
**Expected behavior**
We want what happens when there are still some NAs, but with an error e.g.:
```r
rep_tri <- as_reporting_triangle(data, max_delay = 25)
#> Error The dataframe contains report dates beyond the final reference date.
```

kaitejohnson:
## Description

This PR closes #267. It adds the following:
- a warning within `detect_structure()` if there are no NAs in the reporting triangle, which will error if the user calls `as_reporting_triangle.data.frame()` on a complete reporting triangle. We want this to warn because the user might still want to create this object even if they aren't going to use it for nowcasting -- for example, if they want to use a `reporting_triangle` object that contains complete observations and pass it to `estimate_delay` to get a delay PMF (as this will simplify their pre-processing by avoiding them needing to construct the reporting triangle themselves manually. 
- an error if the user passes in a `reporting_triangle` object with a `reporting_triangle_matrix` with no NAs. This matrix can't be nowcasted because theres nothing to be estimated.
- tests for both cases

## Checklist

## Summary by CodeRabbit

* **Improvements**
  * Added validation to ensure data contains missing observations (NAs) for nowcasting, with an error message if none are present.
  * Added warning when no missing values are detected in the reporting triangle.

* **Documentation**
  * Updated documentation clarifying that missing observations must be coded as NAs rather than 0s.

* **Tests**
  * Added test coverage for validation and warning behavior.

### DIFF
--- a/R/apply_delay.R
+++ b/R/apply_delay.R
@@ -56,8 +56,8 @@ apply_delay <- function(reporting_triangle, delay_pmf) {
     cli_abort(
       message = c("`reporting_triangle` doesn't contain any missing values, there is nothing to nowcast.", # nolint
         "i" = "Check to make sure missing observations are coded as NAs rather than 0s.", # nolint
-        "i" = "If performing nowcasts retrospectively, report times after the last reference date should not be available." # nolint
-        "i" = "See `?construct_triangle` for more details on creating a reporting triangle from a complete set of observations."
+        "i" = "If performing nowcasts retrospectively, report times after the last reference date should not be available.", # nolint
+        "i" = "See `?construct_triangle` for more details on creating a reporting triangle from a complete set of observations." # nolint
       )
     )
   }


R/detect_structure.R
```diff
@@ -29,8 +29,12 @@
 #' detected_structure
 detect_structure <- function(reporting_triangle) {
   n_row_nas <- sum(is.na(rowSums(reporting_triangle)))
-  # Structure is NULL if there are no NAs
+  # Structure is 0 if there are no NAs
   if (n_row_nas == 0) {
+    cli_warn(
```
seabbs: At most this seems like it should be a message. I would advocate for nothing here due to the many other use cases for this data restructure.

R/baselinenowcast.R
```diff
@@ -89,6 +91,14 @@ baselinenowcast.reporting_triangle <- function(
   tri <- data$reporting_triangle_matrix
   output_type <- arg_match(output_type)
   assert_integerish(draws, null.ok = TRUE)
+  n_row_nas <- sum(is.na(rowSums(data$reporting_triangle_matrix)))
+  if (n_row_nas == 0) {
+    cli_abort(
+      message = c("`data$reporting_triangle_matrix` doesn't contain any missing values, there is nothing to nowcast.", # nolint
```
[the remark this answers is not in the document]
kaitejohnson: Have added a (rather verbose) message about checking to make sure that report times after the time of the nowcast have been excluded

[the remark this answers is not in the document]
seabbs: The low level interface is for people to use that is why its exported. This way they will just try to do the same thing but with their own code as they won't know we have any tools for it. 

More generally why is this check here and not in a lower level function its relevant for? If it is in that function does adding this bloat to this function help that much?

tests/testthat/test-as_reporting_triangle.R
```diff
@@ -362,3 +362,20 @@ test_that("assert on reporting triangle works as expected", {
   rep_tri5$structure <- rep(6, ncol(rep_tri$reporting_triangle_matrix))
   expect_error(assert_reporting_triangle(rep_tri5))
 })
+
+test_that("`as_reporting_triangle()` appropriately warns if there is nothing to be nowcasted (no unobserved cases in reporting triangle)", { # nolint
```
seabbs: its info now?

R/detect_structure.R
```diff
@@ -29,8 +29,12 @@
 #' detected_structure
 detect_structure <- function(reporting_triangle) {
   n_row_nas <- sum(is.na(rowSums(reporting_triangle)))
-  # Structure is NULL if there are no NAs
+  # Structure is 0 if there are no NAs
   if (n_row_nas == 0) {
+    cli_alert_info(
```
seabbs: As written this implies its never okay for this to be the case whereas it is actually part of the supported data format. I think I would either just have this check where it matters for functionalit y or reword it so that its clearer what this might mean.

R/baselinenowcast.R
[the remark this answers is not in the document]
seabbs: ?

seabbs:
Waiting on response to q in specific comments.

tests/testthat/test-as_reporting_triangle.R
kaitejohnson: Yes its `cli_alert_info` and so I changed this to say `appropriately messages`

PR #304, opened 2025-11-06
Fix #261: Extract time alignment logic into helper function

### LINKED ISSUE #261: Make time alignment a helper function
Current implementation in `new_baselinenowcast_df`:
```r
 spine_df <- data.frame(
    reference_date = reference_dates,
    time = seq_along(reference_dates)
  )

  baselinenowcast_df_dates <- merge(baselinenowcast_df,
    spine_df,
    by = "time",
    all.x = TRUE
  )
```

The proposed solution is to write a helper function that indexes the vector of reference dates and then joins them to the time indices in the baselinenowcast_df (the output of `sample_nowcasts()` which does not contain any dates). 

this looks like it should probably be a helper function at some point

_Originally posted by @seabbs in https://github.com/epinowcast/baselinenowcast/pull/242#discussion_r2435202976_

seabbs:
**This is entirely from an agent but I have manually reviewed and it looks good to me**

## Summary
Refactors time-to-date alignment logic from `new_baselinenowcast_df()` by extracting lines 49-58 into a new internal helper function `.align_time_to_dates()`.

## Changes
- Created `.align_time_to_dates()` helper function to handle merging reference dates with integer time indices
- Updated `new_baselinenowcast_df()` to call the new helper instead of containing alignment logic inline
- Placed helper function after its first use following project convention (after `new_baselinenowcast_df()`)
- Added two unit tests verifying correct behaviour:
  - Sequential time indices merge correctly with dates
  - Non-sequential time indices handled properly

## Rationale
This refactoring improves code organisation by:
- Isolating time alignment logic from dataframe construction concerns
- Making the codebase easier to understand and maintain
- Following separation of concerns principle
- Maintaining consistency with project patterns for helper functions

## Testing
- All 631 existing tests continue to pass
- Added 2 new unit tests for `.align_time_to_dates()` helper function
- No breaking changes to public API
- Function behaviour remains identical

## Related Issues
Fixes #261

## Summary by CodeRabbit

* **Refactor**
  * Improved internal code organization for baseline nowcast data processing.

* **Tests**
  * Added unit tests to validate data alignment and date mapping functionality.


R/baselinenowcast_df-class.R
```diff
@@ -72,6 +66,35 @@ new_baselinenowcast_df <- function(baselinenowcast_df,
   return(result)
 }
 
+#' Align integer time indices with reference dates
+#' @description Internal helper function that merges reference dates into a
+#'   nowcast dataframe, replacing integer time indices with actual dates.
+#'
+#' @param baselinenowcast_df Data.frame containing nowcast information with an
```
kaitejohnson: For a separate PR I think but am wondering if we should change the input of `new_baselinenowcast_df` so that the input isn't a `baselinenowcast_df` but maybe a `raw_nowcast_df` (e.g. the output of `sample_nowcasts()`) so as not to confuse with the class object. I can make a separate issue.

Issue #148, opened 2025-05-16
Consider adding a message when the delay_pmf passed in isn't a simplex


kaitejohnson:
We might want this to be flexible, but if its a mistake in a separate estimation of the delay we probably want to warn the user. 

https://github.com/epinowcast/baselinenowcast/pull/140/files#r2093174804

PR #308, opened 2025-11-06
Relax CDF/PMF assumptions to support negative corrections

### LINKED ISSUE #278: Relax assumptions that the CDF is monotonically increasing/the PMF sums to one and test using a synthetic example
**Is your feature request related to a problem? Please describe.**
Currently, I think the validation checks in several places enforce an assumption that the cdf is monotoncially increasing/the pmf sums to one. 
If we drop these assumptions it allows for more flexible modelling as we can allow for positive and negative corrections. 
More generally we want to pair this with observation models that allow negative values in order to make this feature robust to small counts.

**Describe the solution you'd like**
Drop checks on the PMF and/or move them to warns or info statements. Add an example using dummy data of upwards and downward corrections and make sure this works (this should be the first thing to do as it will help when stepping through the code to identify any current checks). Ideally, this would be extended into a vignette.
The only risk I see with this is that if people don't actually have downwards corrections then this change (if it is one) could lead to them having silent errors. I think as long as we keep having a message then the risk here is very slow.

**Describe alternatives you've considered**
If we have none of these checks we just need an example walking through this use case.

**Additional context**
This is a major advantage of these non-generative models and so one we should clearly support. It also matches the general aim of the package of being "hands free/context free" for users who are looking for this. I think this could be written into a small paper quite easily if that was a key selling point of adding this feature and walking through the checks etc. My estimate of dev time on this for the complete feature with examples in a vignette is less than a week.

seabbs:
**This is from an agent but I have reviewed. I think it is at a stage where it is worth looking at from a high level but as there is a lot here I would expect some potential issues. I am still thinking a bit about the implications for the maths in estimate_delay, apply_delay etc. but on the face of it I think this is reasonable. Note the `preprocess` pattern is there to allow for extensibility.**

## Summary

This PR relaxes validation assumptions that enforce monotonically increasing CDFs and PMFs summing to one, enabling more flexible modelling with both positive and negative corrections.

Changes include:
- Added `preprocess` parameter to `estimate_delay()` to control negative value preprocessing
- Exported `preprocess_negative_values()` function for user control
- Improved PMF validation messages to be informative rather than restrictive
- Added negative value safety checks to `fit_nb()` for small count robustness
- Created comprehensive test suite with synthetic negative PMF example data
- Updated documentation to reflect new functionality

## Test Plan

- All existing tests pass (684 tests, 0 warnings)
- Added new test file `test-negative-pmf-workflow.R` with 32 tests validating negative PMF handling
- Enhanced tests in `test-apply_delay.R`, `test-estimate_delay.R`, and `test-estimate_uncertainty.R`
- Example dataset `example_negative_pmf` (data) demonstrates upward and downward corrections

## Related Issues

Closes #278
Partially addresses #148 (negative value handling)

## Release Notes

* **New Features**
  - Added optional preprocessing parameter to control how negative values are handled during delay estimation; defaults to redistributing negatives to earlier delays.
  - Added example dataset demonstrating downward corrections with negative PMF values.

* **Improvements**
  - Enhanced input validation with clearer error messages for negative observations and predictions.
  - Improved PMF validation messaging to display actual sum values.

* **Removed**
  - Removed `baselinenowcast.reporting_triangle()` and `as_reporting_triangle()` public functions.


seabbs:
[@coderabbitai](https://github.com/coderabbitai) review

seabbs:
### Additional Note: Limitation on Negative Final Counts

The validation requiring `delay_pmf[1] >= 0` means this implementation **does not support cases where the final total count for a reference time could be negative**.

Since the first PMF entry represents the baseline reporting pattern (delay = 0), requiring it to be non-negative ensures that initial observations are positive. Combined with the iterative nowcasting formula, this prevents scenarios where the estimated total count `N_total` becomes negative.

**Practical Implication**: This limitation means we cannot handle data where systematic downward corrections are so large that they result in negative final counts for a reference time (e.g., more cases are reclassified away than were initially reported).

This is likely acceptable because:
- Such scenarios would be extremely rare in practice
- They would suggest fundamental data quality issues rather than routine corrections
- The use case focuses on small systematic adjustments, not large-scale reclassifications

If such data is encountered, users should consider:
1. Using the default preprocessing to redistribute negative values
2. Aggregating to a coarser temporal resolution where corrections balance out
3. Investigating upstream data quality issues

kaitejohnson:
Thanks for taking this on. I think I want to try out a few more test cases, as my concern is that `.calc_modified_expectation` might return a negative value if `delay_cdf_prev` is much greater than 1 for a low value of the `delay_pmf`... Therefore we might need to put some guardrails on the delay cdf whenever `preprocess` is NULL? 

I also think it would be good to test an example with some aggregated summary metric just as a test (with say a normal observation model) as this would be a very common epi use case for this.

seabbs:
> I also think it would be good to test an example with some aggregated summary metric just as a test (with say a normal observation model) as this would be a very common epi use case for this.

As this needs a new observation model I think that is really another issue/PR. You can still get away with a count observation model here as long as your aggregated target is positive which in most settings will be the case

kaitejohnson:
Fair on both points... Still, it might be good to have a coherent error message if expected N is ever computed to be negative as this would at least tell users/ those in real settings what is causing the issue.

kaitejohnson (approved):
@seabbs Look over my changes but addressed merge conflicts, added preprocess as an arg, and added a very light test...

PR #345, opened 2025-12-03
Rename construct_triangle functions to apply_reporting_structure

### LINKED ISSUE #335: Construct triangles: Do we need it?
As the title. At the moment all it does is validate that the input is a list and then uses lapply with nothing else. Is there much point to having this vs just letting people write lapply(their_list, construct_triangles)?

If removing we likely want to do so soon as then we don't need to depreciate it

seabbs:
## Summary

- Renamed `construct_triangle()` to `apply_reporting_structure()`
- Renamed `construct_triangles()` to `apply_reporting_structures()`
- Updated all references throughout codebase including documentation, tests, and vignettes

The new names better describe what these functions do: they apply a reporting structure (triangular or ragged) to truncated reporting triangles to generate retrospective data.

Closes #335

## Test plan

- [x] All existing tests pass
- [x] Documentation updated with `devtools::document()`
- [x] All references to old function names updated
- [x] Examples in documentation still work

## Summary by CodeRabbit

* **Refactor**
  * Two public functions have been renamed: `construct_triangle` is now `apply_reporting_structure`, and `construct_triangles` is now `apply_reporting_structures`. Function behavior and signatures remain unchanged. Code using these functions will require updates to use the new names.

* **Documentation**
  * All documentation, user guides, examples, and vignettes have been updated throughout the package to reflect the renamed functions.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/R/apply_reporting_structure.R
+++ b/R/apply_reporting_structure.R
@@ -45,7 +45,8 @@ apply_reporting_structures <- function(truncated_reporting_triangles,
   return(reporting_triangles)
 }
 
-#' Apply reporting structure to generate a single retrospective reporting triangle
+#' Apply reporting structure to generate a single retrospective reporting
+#'   triangle
 #'
 #' This function applies a reporting structure to a truncated reporting triangle
 #'   by setting observations to NA row by row from the bottom up based on the
--- a/R/estimate_and_apply_uncertainty.R
+++ b/R/estimate_and_apply_uncertainty.R
@@ -14,8 +14,9 @@
 #' To obtain estimates of uncertainty parameters, use
 #'   [estimate_uncertainty_retro()]. For full control over individual steps
 #'   (e.g., custom matrix preparation, alternative aggregation), use the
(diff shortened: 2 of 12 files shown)


R/apply_reporting_structure.R
```diff
@@ -18,15 +19,15 @@
...
 #' # With custom structure
-#' retro_rts_custom <- construct_triangles(
+#' retro_rts_custom <- apply_reporting_structures(
 #'   trunc_rts,
 #'   structure = 2
 #' )
 #' retro_rts_custom
-construct_triangles <- function(truncated_reporting_triangles,
+apply_reporting_structures <- function(truncated_reporting_triangles,
                                 structure = 1,
```
kaitejohnson: Maybe out of scope for this PR but should we consider calling this `reporting_structure` rather than `structure`? Only ask because `structure` does have another meaning in R and it would connect this function's name and its arguments.

seabbs: Closing this but please do open an issue for discussion

PR #356, opened 2025-12-17
Issue 354: Fix `example_reporting_triangle` 

### LINKED ISSUE #354: `example_reporting_triangle` should probably be estimable with main entry function
**Describe the bug**
1. The current `example_reporting_triangle` provided as package data doesn't meet our minimum requirements for a dataset to produce a probabilistic nowcast. We should just make this larger to ensure that it does
2. Even setting `output_type = "point"` still returns this error message, which is problematic as for the point estimate we shouldn't need additional rows.

**To Reproduce**
```r
#' baselinenowcast(example_reporting_triangle)
#' Error in `.handle_target_exceeds_avail()` at baselinenowcast/R/allocate_reference_times.R:182:5:
#' ! 5 reference times available and 7 are needed, 5 for delay estimation and 2 for uncertainty estimation.
#' ✖ Probabilistic nowcasts cannot be generated.
#' Run `rlang::last_trace()` to see where the error occurred.
```
And
```r
#' baselinenowcast(example_reporting_triangle, output_type = "point")
#' Error in `.handle_target_exceeds_avail()` at baselinenowcast/R/allocate_reference_times.R:182:5:
#' ! 5 reference times available and 7 are needed, 5 for delay estimation and 2 for uncertainty estimation.
#' ✖ Probabilistic nowcasts cannot be generated.
#' Run `rlang::last_trace()` to see where the error occurred.
```

**Expected behavior**
1. The example data should be able to produce a probabilistic nowcast, even if there are warnings about a different number of available reference times than expected
2. If just producing a point forecast, it shouldn't matter that there aren't enough reference times for uncertainty estimation, and a point nowcast should be produced.

kaitejohnson:
## Description

This PR closes #354. It edits the `baselinenowcast.reporting_triangle()` so that when the user specifies a point nowcast, the minimum number of retrospective nowcasts is set to 0. This allows a point nowcast to be produced even if there is only one complete row of observed delays. 

It also edits the `example_reporting_triangle` so that it can produce a probabilistic nowcast from `baselinenowcast` without erroring. 

## Checklist

## Summary by CodeRabbit

* **Bug Fixes**
  * Refined uncertainty estimation for nowcasts to correctly account for output type—point forecasts and probabilistic forecasts now use appropriate configuration settings.

* **Documentation**
  * Extended example data with additional reference times for demonstration purposes.

<sub>✏️ Tip: You can customize this high-level summary in your review settings.</sub>

### DIFF
--- a/R/baselinenowcast.R
+++ b/R/baselinenowcast.R
@@ -349,7 +349,9 @@ baselinenowcast.data.frame <- function(
       data = data_clean,
       strata_cols = strata_cols
     )
-    pooled_triangle <- as_reporting_triangle(pooled_df)
+    pooled_triangle <- as_reporting_triangle(pooled_df,
+      delays_unit = delays_unit
+    )
 
     # Apply preprocessing if provided
     processed_pooled_triangle <- pooled_triangle
--- a/tests/testthat/test-baselinenowcast-dataframe.R
+++ b/tests/testthat/test-baselinenowcast-dataframe.R
@@ -540,7 +540,7 @@ test_that(paste0(
       geom_line(aes(
         x = reference_date, y = pred_count, group = group_var,
         color = type
-      ), size = 0.1, alpha = 0.5) +
+      ), linewidth = 0.1, alpha = 0.5) +
       geom_point(aes(x = reference_date, y = cases)) +
       facet_wrap(~age_group, nrow = 3, scales = "free_y") +
       theme_bw()


kaitejohnson:
> If making this change the automagical switch needs to be in the user facing documentaiton. For now I might prefer the more minimal version which just changes example data and then look at changing the default down the line?

That's fair -- I guess the user can still produce a point nowcast from the minimal reporting triangle, they just can't do it from the wrapper function `baselinenowcast`. My slight preference would be to keep as is in this PR so they can use `baselinenowcast` with `output_type = "point"` but that we add the documentation that specifies that for point nowcasts, the minimum data requirement is just the minimum requirement for delay estimation, since the buffer of 2 retrospective nowcast times isn't used in that instance. It seems problematic to return an error about their being insufficient reference times for retrospective nowcasts for uncertainty estimation when the user isn't estimating the uncertainty.

PR #362, opened 2026-01-05
Issue 363: truncate triangle to pooled triangle maximum delay

### LINKED ISSUE #363: `strata_sharing` option in `baselinenowcast.dataframe` requires pre-processed grid expansion and identical `max_delay`
As described in the issue title, `baselinenowcast.dataframe(strata_sharing = c("delay", "uncertainty)` requires that the user has already expanded the grid in their data.frame to all combinations of reference and report dates (even if 0 cases are observed) across all strata and that the maximum delays are the same across all strata. 

Note: We do not require the dataframe to be pre-processed with grid expansion for the non-strata sharing option, we do it within the `as_reporting_triangle` operation 

Options:
- fix this in `combine_triangles_df` and `baselinenowcast.dataframe` (see #362 )
- ignore this for now, and fix it in #360 by ensuring that all strata have the same maximum delay and all are expanded

I think I am slightly in favor of implementing the the fix in #362 so that the functionality works and then we will use some of that logic in #360. Then we can have tests for the `.data.frame` method that can be used for the `.reporting_triangle_df` method....

kaitejohnson:
This PR closes #363.
It is still a WIP as we need to discuss whether implementing it here makes sense. 

## Description

This PR ensures three things, two changes to `.combine_triangles_df` and one to `baselinenowcast.dataframe`.
- `.combine_triangles_df` creates a data.frame with a maximum delay that is the minimum of the maximum delays in the data across the strata passed in. 
- `.combine_triangles_df` ingests dataframes indexed by reference and report dates for each strata and does an expand grid to fill in 0s on counts that are not reported
- `baselinenowcast.dataframe` truncates the `reporting_triangle` being nowcasted to align with the `pooled_triangle`

Noting that the second and third would not be required when we create `as_reporting_triangle_df` functionality as we can mandate that 1. they all have the same max_delay and 2. the grids are expanded within all strata

I do think in order to achieve 1. in `as_reporting_triangle_df` we will need some of the logic for finding the minimum of the maximum delays across strata... 

There are arguments for ignoring this entirely and focusing on `as_reporting_triangle_df` functionality instead and I am open to that. 

Still to do:
- add tests for `.combine_triangles_df` and `baselinenowcast.dataframe` to check both of these examples work

## Checklist

### DIFF
--- a/R/baselinenowcast.R
+++ b/R/baselinenowcast.R
@@ -351,6 +351,7 @@ baselinenowcast.data.frame <- function(
   if (all(strata_sharing != "none")) {
     pooled_df <- .combine_triangle_dfs(
       data = data_clean,
+      delays_unit = delays_unit,
       strata_cols = strata_cols
     )
     pooled_triangle <- as_reporting_triangle(pooled_df,
@@ -390,7 +391,7 @@ baselinenowcast.data.frame <- function(
     }
   }
 
-  # Nowcast on each reporting triangle and bind into a long data.frame
+  # Truncate reporting triangle to the max delay of the pooled triangle
   list_of_rep_tris_compatible <- lapply(
     list_of_rep_tris,
     function(x) {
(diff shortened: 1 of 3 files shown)


seabbs:
Is there an issue for this? There is an arguement this should error with a warning and the user should fix it in preprocessing. there are choices to be made about this in the new df data type as that could enforce that all strata have the same D and so avoid this.

PR #427, opened 2026-03-25
Issue 426: add sentence about how 0-day delays are implicitly handled

### LINKED ISSUE #426: Clarify NSSP preprocessing language
@lauraajones2 has checked and confirmed that the NSSP records with negative delays of greater than a day are likely due to data entry errors, but that those that occur on the same day should be treated as 0-day delays. This does not change the code but the language should be changed to reflect that this is the expert opinion of those closest to the data (and that this assumption should not be made broadly, but that the cause of negative delays should be investigated within the data).

kaitejohnson:
## Description

This PR closes #426. After discussing with @lauraajones2 we decided it would be better to be explicit about how this code is handling within the same day negative delays. In this case, we decided that we want to treat all negative delays that are less than 24 hours as 0-day delays, and all negative delays greater than 24 hours we want to exclude. 

This means that for delays between 0 and -1, we want to manually set these as having the same reference and report date. This prevents us from excluding the edge case when a patients report date is 11 pm one day and their visit start date is labeled as 1 am the next day. In this case, we will treat this as a visit date and a report date both of which occurred on the next day. 

This threshold for which we consider negative delays to be 0-day delays could be any number of hours e.g. 6, 12 etc. 

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Updated the nowcast vignette with refined methodology for managing negative diagnosis-report delays. New approach includes filtering delays exceeding 24 hours negative, adjusting mild negative delays to same-day reporting, and improved narrative documentation. Changes clarify the impact on case count aggregation and data processing.

### DIFF
--- a/vignettes/nssp_nowcast.Rmd
+++ b/vignettes/nssp_nowcast.Rmd
@@ -178,8 +178,10 @@ The choice of how to handle negative delays should be guided by the data experts
 
 For nowcasting, we want to compute the number of incident cases indexed by reference and report date, so we can aggregate by reference and report date and compute the delay distribution.
 ```{r}
-count_df <- clean_line_list_filtered |>
-  count(reference_date, report_date, name = "count")
+count_df <- count(clean_line_list_filtered,
+  reference_date, report_date,
+  name = "count"
+)
 head(count_df)
 ```
 
--- a/vignettes/nssp_nowcast.Rmd
+++ b/vignettes/nssp_nowcast.Rmd
@@ -157,9 +157,12 @@ head(clean_line_list)
 ```
 ## Remove cases where the diagnoses code was reported before the visit start date by more than 24 hours
 
-Based on our understanding of the data generating process for this particular dataset, we believe that negative delays (instances where the time stamp of diagnosis-code reporting comes before the time stamp of the visit start date) with a magnitude of less than 24 hours likely reflects near synchronous recording and should be treated as a 0-day delay.
-In this case, we will set the report date to be the same day as the reference date (a 0-day delay).
-Whereas negative delays of greater than 24 hours we believe to reflect data entry errors, therefore, for the purpose of nowcasting and estimating the delay distribution, we will exclude these cases.
+Looking at this data, we can see that there is one case where there is a negative delay, which indicates that the time stamp of the diagnosis update was recorded before the start of the visit.
+Depending on the way that the data is generated, this could be a true negative value, for example if the patient had previously been tested elsewhere before arriving at the ED.
+Based on our understanding of the data generating process, we believe that negative delays with a magnitude of less than 24 hours likely reflects near simultaneous recording and should be treated as a 0-day delay.
+The 24-hour threshold here is specific to this dataset and should be investigated and set by the user.
+For negative delays of less than 24 hours, we will set the report date to be the same day as the reference date (a 0-day delay).
+Negative delays of greater than 24 hours we believe to reflect data entry errors, since in this dataset patient interactions always start in the ED, therefore, for the purpose of nowcasting and estimating the delay distribution, we will exclude these cases.
 
 ```{r}
 threshold_to_exclude <- -1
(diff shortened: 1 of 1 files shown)


seabbs:
Is this still in draft or ready for review? Sorry it got screened by my notification settings

kaitejohnson:
> Is this still in draft or ready for review? Sorry it got screened by my notification settings

Sorry, didn't realise it was in draft. Ready for review!

vignettes/nssp_nowcast.Rmd
```diff
@@ -155,22 +155,33 @@ clean_line_list <- first_bar_diagnosis |>
   )
 head(clean_line_list)
 ```
+## Remove cases where the diagnoses code was reported before the visit start date by more than 24 hours
 
-## Obtain counts of cases by reference date (visit date) and report date (time of first diagnosis)
+Based on our understanding of the data generating process for this particular dataset, we believe that negative delays (instances where the time stamp of diagnosis-code reporting comes before the time stamp of the visit start date) with a magnitude of less than 24 hours likely reflects near synchronous recording and should be treated as a 0-day delay.
```
seabbs: Quite verbose imo would cut "particular" and similar.

vignettes/nssp_nowcast.Rmd
```diff
@@ -155,22 +155,33 @@ clean_line_list <- first_bar_diagnosis |>
...
+In this case, we will set the report date to be the same day as the reference date (a 0-day delay).
+Whereas negative delays of greater than 24 hours we believe to reflect data entry errors, therefore, for the purpose of nowcasting and estimating the delay distribution, we will exclude these cases.
 
-For nowcasting, we want to compute the number of incident cases indexed by reference and report date, so we can aggregate by reference and report date and compute the delay distribution.
 ```{r}
-count_df_raw <- clean_line_list |>
-  count(reference_date, report_date, name = "count") |>
-  mutate(delay = as.integer(report_date - reference_date))
+threshold_to_exclude <- -1
```
seabbs, line 168: I think it might be quite important to print or otherwise show how many are being excluded and how many are being rewritten by this step.

vignettes/nssp_nowcast.Rmd
```diff
@@ -155,22 +155,33 @@ clean_line_list <- first_bar_diagnosis |>
...
+clean_line_list_filtered <- clean_line_list |>
+  filter(arrival_to_update_delay >= threshold_to_exclude) |>
+  mutate(report_date = ifelse(arrival_to_update_delay < 0,
+    reference_date,
+    report_date
+  ))
 ```
 
-Looking at this data, we can see that there is one case where there is a negative delay, which indicates that the time stamp of the diagnosis update was recorded before the start of the visit.
-Depending on the way that the data is generated, this could be a true negative value, for example if the patient had previously been tested elsewhere before arriving at the ED.
```
seabbs, line 169: You no longer have the nice context for why this might happen and what it means in the data set.

vignettes/nssp_nowcast.Rmd
seabbs: Why 24 hours mention this should be investigated and set by users

vignettes/nssp_nowcast.Rmd
```diff
@@ -155,22 +155,33 @@ clean_line_list <- first_bar_diagnosis |>
...
-In this case, patients interactions with the system always start in the ED, so this is likely due to a data entry error.
-For this reason, we will will choose to exclude all the negative valued delays, however, the choice of how to handle these should be guided be guided by the data experts and their understanding of the most likely reason for the observation to prevent introducing additional bias in this choice.
+In this synthetic dataset, this removed a case where the report date was reported to be greater than 7 days before the visit start date.
+The choice of how to handle negative delays should be guided by the data experts and their understanding of the most likely reason for the observation to prevent introducing additional bias in this choice.
```
seabbs, line 187: Suggest other viable strats i.e all negs removed all negs set to 0 etc?

vignettes/nssp_nowcast.Rmd
```diff
@@ -155,22 +155,33 @@ clean_line_list <- first_bar_diagnosis |>
...
+threshold_to_exclude <- -1
+clean_line_list_filtered <- clean_line_list |>
+  filter(arrival_to_update_delay >= threshold_to_exclude) |>
+  mutate(report_date = ifelse(arrival_to_update_delay < 0,
+    reference_date,
+    report_date
+  ))
 ```
 
-Looking at this data, we can see that there is one case where there is a negative delay, which indicates that the time stamp of the diagnosis update was recorded before the start of the visit.
```
seabbs, line 168: this looking bit is now missing.

seabbs (dismissed):
I think this looks good but have lost some of the nice exploratory detail and I think it needs more guardrails for users.

vignettes/nssp_nowcast.Rmd
kaitejohnson: So for only 25 entries for the line list data, this just excludes the -7 day delay case. Can print that. Would have a larger effect on a real large dataset ofc

kaitejohnson: Nothing getting rewritten here

vignettes/nssp_nowcast.Rmd
kaitejohnson: Yes ok good point will add these as other options

vignettes/nssp_nowcast.Rmd
seabbs: yeah the point is if people run this on their own code we don't want it to do stuff silently without their buy in

vignettes/nssp_nowcast.Rmd
kaitejohnson: Have added back in

vignettes/nssp_nowcast.Rmd
kaitejohnson: Added

vignettes/nssp_nowcast.Rmd
kaitejohnson: Added back in

seabbs (approved):

PR #472, opened 2026-06-24
Issue 470: add time zone to vignette

### LINKED ISSUE #470: Update NSSP vignette to reflect proper handling of time zones
We have received updated guidance from the NSSP admin team about the time zone of date fields used in the NSSP vignette. They have indicated that DischargeDiagnosisMDTUpdates is composed of the successive direct input from the HL7 messages. Therefore it would depend on the submitter (i.e., facility) whether they were in UTC. For most jurisdictions, it's likely that this is mostly in local time, but that may depend on their local facility onboarding/implementation. 

The vignette should be updated to specify time zone appropriately for both C_Visit_Date_Time and DischargeDiagnosisMDTUpdates and flag it as a implementer decision point.

kaitejohnson:
## Description

This PR closes #470. @lauraajones2 Flagged the issue that time zones may be handled incorrectly if not specified directly, since some of the time stamps might go through the NSSP system and be converted to UTC while others are not necessarily. This likely varies by jurisdiction, but to be safe we should encourage users to specify this as variable and convert the two time stamp variables using the specified time zone. 

## Checklist

## Summary by CodeRabbit

* **Documentation**
  * Clarified time zone handling guidance in the vignette with a user-configurable example.
  * Added notes explaining why consistent time zone settings matter when working with timestamps.

* **Bug Fixes**
  * Improved timestamp processing so related time values are interpreted in the same time zone before delay calculations, helping avoid reporting discrepancies.


lauraajones2:
Very minor edits to the first sentence - I think specifying the direct inputs from HL7 messages is key to any local decision about time zones, so for the first sentence I would instead say: "The fields, `DischargeDiagnosisMDTUpdates` and `C_Visit_Date_Time` are composed of direct input from the HL7 messages sent by facilities to NSSP, thus, these updates usually reflect the facility's time zone choice (typically local time, but dependent on local implementation).

seabbs (approved):

kaitejohnson:
@lauraajones2 What does HL7 messages mean? 

@seabbs I will open a separate PR to make these changes

lauraajones2:
@kaitejohnson HL7 messages are standardized strings of text used by facilities to exchange clinical data. It's an important concept for anyone using line-level NSSP data because (usually) an HL7 message comes through for every change to the visit's EHR, and then each NSSP record is continuously updated based off of the hundreds of HL7 messages received for an encounter over time. Jurisdictions onboards facilities to their syndromic program with similar but not identical guidance, so there can be local nuances in how HL7 messages are both sent and subsequently processed (e.g., some jurisdictions may request facilities to stop sending updates to messages after 6 months have passed, some may request they use UTC time for all time stamps, etc.)

kaitejohnson:
@lauraajones2 Do you think it is critical that we use the term HL7 messages or could I paraphrase what you've described here? Otherwise, think we should definitely define it which is fine!

lauraajones2:
For brevity, we can probably use my suggested edits sans mention of HL7 specifically!

"The fields, DischargeDiagnosisMDTUpdates and C_Visit_Date_Time are composed of direct input from the messages sent by facilities to NSSP, thus, these updates usually reflect the facility's time zone choice (typically local time, but dependent on local implementation)."

